PR #19, opened 2022-05-13
Feat/sync to legacy

andreasgan:


cms/src/hooks/sync-to-legacy/filters/assets.ts
```diff
@@ -0,0 +1,107 @@
...
+    } else {
+        patch.EncodingVersion = EncodingVersion.AMS
+    }
+
+    console.log("inserting video", patch)
+    let legacyAsset = await oldKnex<VideoEntity>("video").insert(patch).returning("*")
+    console.log("legacyAsset", legacyAsset)
+    p.legacy_id = legacyAsset[0].Id
+    console.log("p", p)
+    return p
```
KillerX, line 50: Sometimes you do this and sometimes not. Is there a difference?

andreasgan:
missing:
- tag name update
- sections and collections but those can wait - theyre not necessary for MVP

cms/src/hooks/sync-to-legacy/filters/episodes.ts
```diff
@@ -0,0 +1,143 @@
...
+    }
+    let image = null
+    if (p.image_file_id != null) {
+        image = (await c.database("directus_files").select("*").where("id", p.image_file_id))[0];
+    }
+    console.log("directus", p)
+
+    // update it in original 
+    let patch: Partial<EpisodeEntity> = {
+        VideoId: asset?.legacy_id ?? 1041, // TODO: remove this temp id
```
KillerX, line 26: TODO?

cms/src/hooks/sync-to-legacy/filters/episodes.ts
```diff
@@ -0,0 +1,143 @@
...
+        AvailableTo: p.available_to as unknown as Date,
+        AvailableFrom: p.available_from as unknown as Date,
+        LastUpdate: new Date()
+    }
+    if (p.asset_id) {
+        let asset = (await c.database("assets").select("*").where("id", p.asset_id))[0];
+        patch.VideoId = asset.legacy_id
+    } else if (p.asset_id === null) {
+        patch.VideoId = null
+    }
```
KillerX, line 90: Is it intentional that you check that strictly for `null` or should it just be a plain else (not sure what the implications are)

cms/src/hooks/sync-to-legacy/filters/lists.ts
```diff
@@ -0,0 +1,58 @@
+import { oldKnex } from "../oldKnex";
+import { createLocalizable, getStatusFromNew, isObjectUseless, upsertLS } from "../utils";
+import episodes from '../../../btv';
+import { LanguageEntity, CategoryEntity, SeriesEntity } from "@/Database";
+import { ItemsService } from "directus";
+
+export async function createList(p, m, c) {
+    console.log('Item created!');
+    console.log(p,m,c);
```
KillerX, line 9: In general I feel that there is a log of unfiltered logging. Maybe this should be cleaned up (and named correctly) to aid potential debugging.

cms/src/hooks/sync-to-legacy/filters/lists_relations.ts
```diff
@@ -0,0 +1,93 @@
...
+        return;
+    }
+}
+
+
+export async function updateListRelation(p, m, c) {
+    if (m.collection != "lists_relations") {
+        return
+    }
+};
```
KillerX, line 58: If this is used anywhere you might have a bug.

cms/src/hooks/sync-to-legacy/utils.ts
```diff
@@ -12,4 +16,68 @@ export function getStatusFromNew(status: string): number {
...
+        ...e,
+        ...fields
+    }
+}
+
+export function sleep(ms) {
+    return new Promise(resolve => setTimeout(resolve, ms));
+}
+
+export function isObjectUseless(patch: object) {
```
KillerX, line 76: :1st_place_medal: for the naming :D

KillerX (commented):
Some comments on the more general things, I am assuming that the data logic is sane, since it's hard to say that just from a review

cms/src/hooks/sync-to-legacy/filters/assets.ts
andreasgan: Do what? I do p.legacy_id = legacyAsset[0].Id on **create**.

cms/src/hooks/sync-to-legacy/filters/episodes.ts
andreasgan: thanks 👍

cms/src/hooks/sync-to-legacy/filters/episodes.ts
andreasgan: intentional. This is an update filter, so if its "undefined" it wasnt changed, but if its "null" we know directus explicitly set it to null.

cms/src/hooks/sync-to-legacy/filters/assets.ts
KillerX: `return p`

andreasgan: it works without it too, so probably doesnt do anything

PR #21, opened 2022-06-13
Add JWT middleware and `me` query

KillerX:
The middleware is applied to all requests. If the `Authorization` header is not set the request is marked as Anonymous.
The middleware is not responsible for preventing access, it is only supposed to validate the token (if it exists) and extract claims into the context. Claims retrieved from the context later can thus be considered valid and used to Authorization purposes.


backend/graph/schema.graphqls
```diff
@@ -247,6 +247,14 @@ type FAQCategory {
   questions: [FAQ!]!
 }
 
+type User {
+  personId: ID
```
fredrikved, line 251: Do we want this to be called personId? One could argue that this is a bad naming originally in Members...? 

I'd either go with "id" or "userId".

fredrikved: Of course if this is just a "forward" of what we get from auth0, then we don't have much of a choice

fredrikved (approved):

andreasgan:
I'm curious if we need to check hasMembership.
I've never done that
https://github.com/bcc-code/bcc-code.github.io/blob/master/_docs/signon/openid-connect.md

backend/graph/schema.graphqls
KillerX: Yeah, I'll move this to ID, makes more sense, even if members i "wrong".

KillerX:
> I'm curious if we need to check hasMembership. I've never done that https://github.com/bcc-code/bcc-code.github.io/blob/master/_docs/signon/openid-connect.md

I will use this for https://github.com/bcc-code/brunstadtv/pull/21/files#diff-fcaa1411fa7893dd366fed7c9738f2dc0b054503ed682c3df089fbd7946155b0R22 thanks

PR #23, opened 2022-06-15
Added simple script for generating some test data

KillerX:
The script is very basic for now and we should probably improve it as we
need new features. The main goal for now is to create a valid show with
seasons, and episodes in order to be able to start working on
permissions.


andreasgan:
må ikke hardkode secrets matjaz

KillerX:
For the record the "secrets" were an example password and token that are not used anywhere

fredrikved (approved):

PR #24, opened 2022-06-17
Search engine implementation

fredrikved:


andreasgan:
Looks very good to me 👍
About webhooks: Cloud run only allocated CPU during requests, so the webhook endpoint might be buggy when running in cloud run.
So you need to make sure that the webhooks actually get fired (this depends on when directus sends the webhooks, if it's async we might need to use [filter hooks](https://docs.directus.io/extensions/hooks/), or just enable always-on CPU allocation.

backend/directus/events.go
```diff
@@ -0,0 +1,101 @@
...
+	return ""
+}
+
+var itemsEvents = map[string][]func(model string, id int){}
+
+func (handler *EventHandler) On(events []string, callback func(model string, id int)) {
+	for _, event := range events {
+		switch event {
+		case EventItemsUpdate, EventItemsCreate, EventItemsDelete:
+			log.L.Debug().Msgf("Registering Directus webhook-listener for event: \"%s\"", event)
```
KillerX, line 46:
```suggestion
			log.L.Debug().Str("event", event).Msg("Registering Directus webhook-listener")
```
This way we can get event as its own field and can use it for filtering etc later.

Same on line 60 for example

backend/directus/events.go
```diff
@@ -0,0 +1,101 @@
...
+			switch v := entry.(type) {
+			case int:
+				id = v
+			case float64:
+				id = int(v)
+			case string:
+				num, _ := strconv.ParseInt(v, 0, 64)
+				id = int(num)
+			}
+			return id
```
KillerX, line 84: This looks like a nice helper function that will likely come in handy in other places when we talk to DU.

backend/directus/events.go
```diff
@@ -0,0 +1,101 @@
...
+	if model == "" {
+		log.L.Debug().Msg("Collection not supported yet")
+		return
+	}
+
+	var ids []int
+	if !(len(event.Keys) > 0) && event.Key != 0 {
+		ids = []int{event.Key}
+	} else {
+		// Directus returns either floats or strings in the same array... weird?
```
KillerX, line 72: Yeah, it's very geared towards dynamic languages (aka JS)

backend/search/episodes.go
```diff
@@ -0,0 +1,183 @@
...
+
+	err := indexObjects(index, objects)
+	if err != nil {
+		log.L.Error().Err(err).Msg("Failed to index objects")
+		return
+	}
+}
+
+func (service *Service) indexEpisode(item sqlc.Episode) {
+	ctx := context.Background()
```
KillerX, line 152: Any specific reason as to why you are creating a new context here?

backend/base/interfaces.go
```diff
@@ -0,0 +1,14 @@
+package base
+
+type ISearchService interface {
```
KillerX, line 3: No `ISomething`s please.
The interface in go should be defined at the place where it is used, so something like the "base" having an interface makes very little sense."

We can take a closer look at this on Monday.

backend/cmd/api/main.go
```diff
@@ -44,11 +45,30 @@ func playgroundHandler() gin.HandlerFunc {
...
+
+	var err error
+	if a == "" {
+		err = errors.New("missing x-api-key header")
+	} else {
+		err = errors.New("failed to authenticate")
+	}
+
+	c.Status(401)
+	_, _ = c.Writer.WriteString("Unauthorized")
```
KillerX, line 62:
```suggestion
	c.String("Unauthorized")
	c.AbortWithStatus(http.StatusUnauthorized)
```

Gin provides many nice functions like for example https://pkg.go.dev/github.com/gin-gonic/gin#Context.AbortWithStatusJSON

backend/cmd/api/main.go
```diff
@@ -87,6 +108,13 @@ func main() {
 
 	log.L.Debug().Msgf("connect to http://localhost:%s/ for GraphQL playground", config.Port)
 
+	searchService := search.New(config.Algolia.AppId, config.Algolia.ApiKey, db)
+	configureSearchEndpoints(r, config, &searchService)
```
KillerX, line 112: We should talk about this. I have considered doing things this way and my conclusion in the end was that all of endpoint config should happen in the main function. The main reason for it is that when you are looking for it in 6 months you want to have one place to look.

backend/search/query-handler.go
```diff
@@ -0,0 +1,162 @@
...
+	"strings"
+	"time"
+)
+
+type QueryHandler struct {
+	user    any
+	service *Service
+}
+
+func (service *Service) GetQueryHandler(user any) base.ISearchQueryHandler {
```
KillerX, line 19: In go as a rule we do not return interfaces, we accept them.

KillerX (changes requested):
I added some superficial comments, about general go stuff and libraries that we use. I think we need to take a look at the structure of the whole thing in person so we can have a more direct discussion about it.

migrations/000001_schema.up.sql
```diff
@@ -2,10 +2,8 @@
 -- PostgreSQL database dump
```
KillerX, line 2: We need to come up with a better method for coordinating DB changes. I also have a change pending and either merge will result in a huge conflict here. Any clever ideas?

andreasgan: 1 person can be responsible for directus changes (changes to the schema.yml), so any changes must be coordinated with that person.
Anything directus doesn't manage, like db views, can be done via directus migrations.
- This makes the psql dump inconsistent, but just do a new dump during merge.

backend/cmd/api/search.go
```diff
@@ -47,6 +47,23 @@ func searchQueryHandler(client *search.Service) func(*gin.Context) {
...
 }
 
+func searchKeyHandler(client *search.Service) func(*gin.Context) {
+	return func(c *gin.Context) {
+		var query common.SearchQuery
+		// define default options
+		query.Page = 0
+		err := c.BindJSON(&query)
+		if err != nil {
+			log.L.Error().Err(err)
```
KillerX, line 57: IIRC you have to use some form of `.Msg()` in order for the log to actually be logged.

PR #25, opened 2022-06-20
Feat/make create modify non nullable

KillerX:
Update the `date_created` and `date_updated` fields to be non-nullable


KillerX:
This is aimed at the other pull request in order to prevent massive conflicts with the DB migrations

fredrikved (approved):

PR #28, opened 2022-06-22
Use event structure from Directus

fredrikved:
Allow messages from directus to be directly pushed to pub/sub and to the worker.


fredrikved:
I haven't tested this yet, so it's just theoretical until I've set up the pushes.

Is this how it would work though? @KillerX

KillerX (approved):

KillerX (approved):

PR #31, opened 2022-06-24
Feat/crowdin translations

fredrikved:


cms/src/hooks/crowdin/index.ts
```diff
@@ -0,0 +1,28 @@
...
+}
+
+export default ({action, schedule}) => {
+    if (!enabled()) 
+        return
+    
+    action('items.create', updateOrSetTranslationAsync)
+    action('items.update', updateOrSetTranslationAsync)
+
+    schedule('0 * * * *', sync)
```
KillerX, line 22: Do you know how this is triggered. Because if it's in the background, then we could have many instances doing the same thing, or none at all, depending on what is currently running. IMO we should be able to trigger this via an url.

KillerX (commented):
I'm not very happy with the dynamic queries, where we could theoretically get unsanitized input. Some cases could cause issues even by accident, and prevent this from running.

Another thing to note is that we should somehow be able to monitor this jobs. The easiest way is with BetterUptime. That basically involves hitting a url.

KillerX (commented):
Some things to take a look at. I want to test a few things but feel free to shoot back comments.

backend/cmd/jobs/server/server.go
```diff
@@ -71,14 +86,27 @@ func (s server) ProcessMessage(c *gin.Context) {
...
 
+	if _, ok := runOnceEvents[e.Type()]; ok {
+		if messageCache.Contains(e.ID()) {
+			log.L.Debug().Str("MsgId", e.ID()).Msg("ignoring processed message")
+			c.Status(http.StatusOK)
+			return
+		} else {
+			messageCache.Set(e.ID(), true, cache.WithExpiration(time.Minute*5))
+		}
+	}
```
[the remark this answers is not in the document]
fredrikved: Not very important for it to run once, but for example translation sync can take ages. 

Would starting a go routine for that be better? (See newest commit)

backend/crowdin/events.go
```diff
@@ -0,0 +1,106 @@
...
+
+type config interface {
+	GetCrowdinToken() string
+	GetCrowdinProjectIDs() []int
+}
+
+func HandleEvent(ctx context.Context, services services, config config, event cloudevents.Event) (err error) {
+	client := New(config.GetCrowdinToken(), ClientConfig{
+		ProjectIDs: config.GetCrowdinProjectIDs(),
+	})
```
KillerX, line 33: Would it make sense for this to be global? Then you could also potentially use it as a lock to prevent multiple uses on the same instance at least.

backend/crowdin/events.go
```diff
@@ -0,0 +1,106 @@
...
+			return nil, err
+		}
+		return toTranslationSources(ts), nil
+	case "episodes":
+		ts, err := d.ListEpisodeTranslations(ctx, language, false, id)
+		if err != nil {
+			return nil, err
+		}
+		return toTranslationSources(ts), nil
+	}
```
KillerX, line 84:
```suggestion
        var err Error
        var ts <TYPE>
	case "shows":
		ts, err := d.ListShowTranslations(ctx, language, false, id)
	case "seasons":
		ts, err := d.ListSeasonTranslations(ctx, language, false, id)
	case "episodes":
		ts, err := d.ListEpisodeTranslations(ctx, language, false, id)
	}
	if err != nil {
			return nil, err
		}
		return toTranslationSources(ts), nil
```
Maybe?

KillerX (approved):
I think it's ok.
We can do some refactoring down the road as we see the need.

One thing I would like to see is tracing on the top and on any lower level functions you think may be interesting, for the purposes of seeing the performance in prod.

backend/crowdin/events.go
fredrikved: Need the type conversion. If there's a way to do this explicitly/implicitly another way, then that's nice, but I need to convert [](TranslationSource) to []TranslationSource

PR #37, opened 2022-06-29
Execute refresh as the view owner

KillerX:
error: "pq: permission denied for table materialized_views_meta"
https://dba.stackexchange.com/a/171933


KillerX:
I'm merging this in order to continue but let me know if there are any issues that you see.

PR #43, opened 2022-07-04
Implement first POC for the episodes dataloader

KillerX:
TODO:

- [ ] Missing auth check on episode level
- [ ] Some data is missing on episodes
- [ ] How do we expose localization for ease of use on the frontend +
cache

Note: This may not be targeting the correct branch for reasons. :)

### DIFF
--- a/backend/program/episode.go
+++ b/backend/program/episode.go
@@ -9,6 +9,7 @@ import (
 	"github.com/samber/lo"
 )
 
+// NewBatchLoader returns a configured batch loader for GQL Program
 func NewBatchLoader(queries sqlc.Queries) *dataloader.Loader[int, gqlmodel.Program] {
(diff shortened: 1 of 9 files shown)


backend/program/episode.go
```diff
@@ -0,0 +1,56 @@
...
+
+func NewBatchLoader(queries sqlc.Queries) *dataloader.Loader[int, gqlmodel.Program] {
+	batchLoadPrograms := func(ctx context.Context, keys []int) []*dataloader.Result[gqlmodel.Program] {
+		results := []*dataloader.Result[gqlmodel.Program]{}
+
+		ids := lo.Map(keys, func(key int, _ int) int32 {
+			return int32(key)
+		})
+
+		res, err := queries.GetEpisodesWithTrnaslationsByID(ctx, ids)
```
fredrikved, line 20:
```suggestion
		res, err := queries.GetEpisodesWithTranslationsByID(ctx, ids)
```

PR #48, opened 2022-07-07
Chore/cleanup search and variables

fredrikved:


KillerX (approved):
Thanks for being my spellcheck :D

PR #57, opened 2022-07-11
implement endpoint in gql

fredrikved:


backend/graph/schema.graphqls
```diff
@@ -257,6 +265,11 @@ type QueryRoot{
     id: ID!
   ): Section
 
+  search(
+    queryString: String!
+    page: Int!
+  ): SearchResult!
+
```
KillerX, line 272: Fyi: This is the generally accepted way of doing pagination in GQL:
https://graphql.org/learn/pagination/

The idea is that the API doesn't control what a page is.

KillerX (approved):

PR #59, opened 2022-07-12
Updated ingest format to extract audio langs

KillerX:


backend/asset/ingest.go
```diff
@@ -115,6 +117,35 @@ func SafeString(s string) string {
...
+// GetLanguagesFromVideoElement formatted as per AWS:
+// https://aws.amazon.com/blogs/media/smil-using-aws-elemental-mediapackage-vod/
+// For example:
+// ```
+//      <video name="example_1080.mp4" systemLanguage="eng,spa,fra" audioName="English,Spanish,French"/>
+// ```
+// If systemLanguage param is not present the return will be an empty array
+func GetLanguagesFromVideoElement(videoElement smil.Video) []directus.AssetStreamLanguage {
+
+	systemLanuages := []string{}
```
fredrikved, line 129:
```suggestion
	systemLanguages := []string{}
```
minor typo

backend/asset/ingest_test.go
```diff
@@ -42,3 +45,61 @@ func TestCalculateDuration(t *testing.T) {
 		assert.Equal(t, out, a.DurationInS, in)
 	}
 }
+
+func TestGetLanguagesFromVideoElement(t *testing.T) {
+	bytes, err := ioutil.ReadFile("./smil/smil2.xml")
```
fredrikved, line 50: If this non-go file is only used in testing, shouldn't it be in the same directory as the test? Or perhaps another root directory for non-go assets? 
Just a thought

fredrikved (approved):

backend/asset/ingest_test.go
KillerX: Normally yest, but I forgot to commit a line in another test that takes this new file into use as well. I think in this case the location is ok as the file is fed to the the code in the same location as the test file, only the resulting object is actually used in the test in this folder. Meaning if the file changes the code in the same folder as the file will need to change as well.

KillerX: I think I will create a subfolder `testdata` as suggested here: https://medium.com/soon-london/testing-with-golden-files-in-go-7fccc71c43d3

PR #60, opened 2022-07-12
Feat/season and show loaders

fredrikved:
Added several new views for this, generalised access validation and fixed the refresh view function


backend/common/loader.go
```diff
@@ -0,0 +1,104 @@
...
+	"github.com/graph-gophers/dataloader/v7"
+	"github.com/samber/lo"
+)
+
+// NewKeyedListBatchLoader returns a configured batch loader for Lists
+func NewKeyedListBatchLoader[t any](
+	factory func(ctx context.Context, ids []int32) ([]t, error),
+	getKey func(item t) int,
+) *dataloader.Loader[int, []*t] {
+	batchLoadFiles := func(ctx context.Context, keys []int) []*dataloader.Result[[]*t] {
```
KillerX, line 14: The variable name here is probably a bit specific :)

cms/src/migrations/9005-new-views-for-access.ts
```diff
@@ -0,0 +1,255 @@
...
+`
+
+const update_access_function = `
+create or replace function update_access(view character varying) returns boolean
+    language plpgsql
+as
+$$
+DECLARE
+    lr timestamptz;
+BEGIN
```
KillerX, line 191: I think you are missing `SECURITY DEFINER` or something like that here. The reason is that the view can only be refreshed by the user who created it and by having that on the function it is executed as the user who defined it.

KillerX (changes requested):
I think the `SECURITY DEFINER` thing needs to be added otherwise the refresh will not work.

KillerX (approved):

PR #61, opened 2022-07-13
resolve show and season on season and episode

fredrikved:
branch for implementing show -> seasons -> episodes and episode -> season -> show

also added a view here for expanded items, to avoid building duplicate queries (single item loading and list loading (children))


backend/graph/model/episode.go
```diff
@@ -23,13 +24,25 @@ func EpisodeFromSQL(ctx context.Context, row *sqlc.EpisodeExpanded) *Episode {
...
+	}
 
 	episode := &Episode{
 		Chapters:         []*Chapter{}, // Currently not supported
 		ID:               fmt.Sprintf("%d", row.ID),
 		Title:            titleMap.Get(languages),
 		Description:      descriptionMap.Get(languages),
-		ExtraDescription: extraDescriptionMap.Get(languages),
+		ExtraDescription: extraDescription,
+		Season:           season,
```
KillerX, line 45: Does the resolver still run if you assign the object here?

backend/graph/resolver.go
```diff
@@ -63,3 +67,23 @@ func resolverForIntID[t restrictedItem, r any](ctx context.Context, id string, l
 
 	return resolverFor(ctx, int(intID), loader, converter)
 }
+
+func itemsResolverFor[k comparable, t restrictedItem, r any](ctx context.Context, id k, loader *dataloader.Loader[k, []*t], converter func(context.Context, *t) *r) ([]*r, error) {
```
KillerX, line 71: The signatures are starting to resemble C++ :D

KillerX (approved):
I assume you have checked what I wrote in the other comment. Looks nice!

backend/graph/model/episode.go
fredrikved: Yes. Runs as expected, but I could avoid assigning it and instead add SeasonID to Episode, etc, as I only need the ID for the resolvers to work.

KillerX: I think this is perfectly fine then.

PR #62, opened 2022-07-13
Simple pagination

fredrikved:


backend/graph/resolver.go
```diff
@@ -68,22 +68,58 @@ func resolverForIntID[t restrictedItem, r any](ctx context.Context, id string, l
...
 	}
 
-	return utils.MapWithCtx(ctx, lo.Filter(items, func(i *t, _ int) bool {
+	available := lo.Filter(items, func(i *t, _ int) bool {
 		// Validate that user has access
 		return user.ValidateAccess(ctx, *i) == nil
-	}), converter), nil
+	})
+
+	// TODO: Not sure if we can deal with pagination before loading data and validating permissions
```
KillerX, line 93: Probably not because we need to filter by permissions...

KillerX (commented):
I have to take a look at pagination.
IIRC there is a quite good reason for the Relay type pagination and it is also natively supported by many tools. You get a few things in between (Nodes, Edges), but as I recall it was worth it.

fredrikved:
Will wait with this then.

PR #66, opened 2022-07-14
Add mapping from legacy language codes to 2 letter

KillerX:


KillerX:
Forgot that go need fallthrough. Ill convert to a map.

PR #77, opened 2022-07-26
Feat/pages and sections

fredrikved:


backend/graph/schema.resolvers.go
```diff
@@ -53,8 +50,61 @@ func (r *episodeSearchItemResolver) Season(ctx context.Context, obj *gqlmodel.Ep
...
+	if after != nil {
+		sections = lo.Filter(sections, func(s gqlmodel.Section, index int) bool {
+			return index > *first
+		})
+	}
+	if first != nil {
+		sections = lo.Filter(sections, func(s gqlmodel.Section, index int) bool {
+			return index < *first
+		})
+	}
```
KillerX, line 85: I feel like there is a copy-paste bug here with first and after

backend/jsonlogic/filter.go
```diff
@@ -0,0 +1,68 @@
+package jsonlogic
```
KillerX, line 1: We should probably add some unit testing here, also around potential SQL injections.

As you note, for now users can't submit this but we should still make sure it's as safe as possible.

Primary reason for unit tests here from my POV is that we don't want to change the results that are produced because that will change the content that is shown. Or at least we should be very aware that we are changing the SQL output here.

cms/custom/interface-query-builder/.gitignore
```diff
@@ -0,0 +1,3 @@
+.DS_Store
+node_modules
```
KillerX, line 2: This 2 we can probably make sure are in the top level .gitignore.

cms/custom/interface-query-builder/src/query-builder/Filter.vue
```diff
@@ -0,0 +1,164 @@
+<template>
```
KillerX, line 1: Would it make sense to publish this as a npm package somehow?

KillerX (commented):
Looking good so far IMO

backend/jsonlogic/filter_test.go
```diff
@@ -0,0 +1,54 @@
...
+	"fmt"
+	"github.com/stretchr/testify/assert"
+	"io/ioutil"
+	"os"
+	"testing"
+)
+
+func TestConvertToSQL(t *testing.T) {
+	// Open our jsonFile
+	jsonFile, err := os.Open("./testdata/filter.json")
```
KillerX, line 14: If you modify the test files a touch and make them something like:

```json
{
    "in": {}, // <- Json logic object
    "out": "X <> 99" // <- Expected result
}
```
then you can just run through all json files in that dir and we can super easily add new cases.

backend/jsonlogic/filter.go
```diff
@@ -16,6 +19,39 @@ func opToDbOp(operator string) string {
...
+	return result
+}
+
+func getValueFromSource(source any) (string, error) {
+	switch v := source.(type) {
+	case map[string]any:
+		if prop, ok := v["var"]; ok {
+			switch vt := prop.(type) {
+			case string:
+				// enough to avoid sql injection? Columns shouldn't have any spaces anyway heh
```
KillerX, line 37: Maybe we can use https://pkg.go.dev/github.com/lib/pq#QuoteIdentifier and https://pkg.go.dev/github.com/lib/pq#QuoteLiteral ?

fredrikved: Oh that's probably better yes

cms/custom/interface-query-builder/src/query-builder/Filter.vue
fredrikved: We can at least publish the query builder itself as an NPM package when we're happy with it. The directus-specific configuration has to be constructed like this to keep typing, vue capabilities and for directus to correctly register the extension.

For now, it's easier to have it in the same project.

KillerX (changes requested):
There is one comment I would like some clarity on before we merge this but it looks good otherwise IMO.

I didn't do a very deep review of the VUE parts as my VUE experience is gluing together obsolete snippets of code from random places. If you think we need a better review feel free to tag Andreas or maybe even Gerard since he probably has the most experience with Directus.

backend/graph/schema.graphqls
```diff
@@ -59,67 +27,86 @@ enum Language{
...
+enum ItemSectionType {
+  cards
+  slider
 }
 
-interface SectionBody {
+type ItemSection implements Section {
   id: ID!
-  items: [Item!]!
+  page: Page! @goField(forceResolver: true)
```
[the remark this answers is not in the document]
fredrikved: At the moment sections are connected to only one page, but could definitely change it for it to be referenced from multiple pages.

Just depends on what a section is, and if it's better for them to just be unique per page. Section in DB is this: 
Title
Description
Type
CollectionId (the source of the content)

Collections can be used by multiple sections.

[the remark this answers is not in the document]
fredrikved: GQL schema does not 100% represent the DB structure here.

KillerX (dismissed):

PR #80, opened 2022-07-27
Run Unit tests in CI

### LINKED ISSUE #72: Add unit test for go into docker build step
should besically just be a `go test ./...` somewhere. I don't really care about coverage but the build should fail the the tests fail.

KillerX:
Fixes #72

### DIFF
--- a/backend/Dockerfile.api
+++ b/backend/Dockerfile.api
@@ -7,6 +7,7 @@ ADD go.sum /go/src/go.sum
 RUN go mod download
 
 ADD ./backend /go/src/backend
+RUN CGO_ENABLED=0 go test ./...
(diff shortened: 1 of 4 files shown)


fredrikved (approved):
We might want to split the tests into a different stage in CI later or? Shouldn't be necessary to have that in the production dockerfile?

KillerX:
> We might want to split the tests into a different stage in CI later or? Shouldn't be necessary to have that in the production dockerfile?

Mnjah, there is the other side: Once we get Staging up we should technically not build a different image for PROD. We should run the exact same image. I'm not 100% sure yet how we will do that but when we do it then I don't see a huge benefit to splitting it.
The tests are not in the final image anyway, since only the binary gets pushed to the final image.

PR #82, opened 2022-07-27
Force tags to be lowercase

KillerX:
Fixes: https://projects.bcc.no/#/tasks/35692468


fredrikved (approved):
As long as the migrations work correctly in our setup (im not convinced that they are run properly...), this should be enough.

KillerX:
I have converted this back to a draft as I need to fix #83 first. As @fredrikved noted the migrations are actually not executed properly.

KillerX:
Never mind. They do run, I was just looking at some tables in truncated mode...

PR #84, opened 2022-07-28
"Id" != "id"

KillerX:
Turns out this was causing errors in Directus and not syncing tags


KillerX:
I'm merging this to validate the changes in DEV.

PR #91, opened 2022-08-09
Added simple implementation of external events

KillerX:
as per https://projects.bcc.no/#/tasks/35687679

There is nothing actually "happening" to the data yet but it can be developed against if needed

### DIFF
--- a/backend/external-events/events.go
+++ b/backend/external-events/events.go
@@ -1,0 +1,121 @@
+package externalevents
+
+import (
+	"encoding/json"
+	"time"
+
+	"github.com/ansel1/merry/v2"
+)
+
+// Based on https://github.com/bcc-code/brunstadtv/blob/ad2e2942eaff5530830fb6a3a3ed5ee449a71cfd/backend/cmd/events/Readme.md
+/*
+{
+	"timestamp": "2009-11-11T08:00:00+09:00", // RFC3339 formatted
+	"source": "some-system", // Arbitrary string identifying the source of the message, Informative
+	"type": "event.happened", // Arbitrary string, identifying the format of the data sent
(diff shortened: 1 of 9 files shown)


backend/external-events/events.go
```diff
@@ -0,0 +1,121 @@
...
+
+// ParseEvent from the json bytes
+func ParseEvent(data []byte) (*Event, error) {
+	e := &Event{}
+	err := json.Unmarshal(data, e)
+	if err != nil {
+		return nil, merry.Wrap(err)
+	}
+
+	if e.Type == TypeSong {
```
fredrikved, line 101: Is a switch case more clean here?

backend/external-events/events.go
```diff
@@ -0,0 +1,121 @@
...
+{
+	"timestamp": "2009-11-11T08:00:00+09:00", // RFC3339 formatted
+	"source": "some-system", // Arbitrary string identifying the source of the message, Informative
+	"type": "event.happened", // Arbitrary string, identifying the format of the data sent
+	"data": {"A": "B", "C": 99 }, // Arbitrary json object containing the actual data
+}
+*/
+
+// Event is the base event wihth data represented as raw json
+type Event struct {
```
fredrikved, line 21: Just a thought.

Maybe "hide" BibleData, SongData, etc behind GetBibleData / GetSongData functions?
We have to check the type whenever we want to retrieve the event data, so might as well wait with the deserialization until we need it, or? That way, we don't need to make pointers, as we know they are not nil.

KillerX: I'll actually wait with this as I don't know how this will be used internally yet.

KillerX: Or, no I'll hide them, Makes it easier to not serialize to json.

PR #93, opened 2022-08-12
Feat/query filter preview

fredrikved:
Simple preview for query filters.


backend/cmd/api/main.go
```diff
@@ -55,6 +61,48 @@ func playgroundHandler() gin.HandlerFunc {
...
+			_ = json.Unmarshal(stringed, &mapped)
+			mapped["type"] = strings.ToLower(t)
+
+			result = append(result, mapped)
+		}
+
+		c.JSON(200, result)
+	}
+}
+
```
KillerX, line 105: Why don't we allow just getting the collection by id via GQL? Can be a new endpoint but I'm kind of reluctant to start adding a REST based api here now.

fredrikved: Because of cache. This endpoint is for internal use.

Could move this to another API for internal use?
Just need to use existing go code.

fredrikved: I don't think we should expose internal endpoints through the same gql schema, but making an admin api could solve our "issues"

backend/cmd/api/main.go
```diff
@@ -61,7 +61,21 @@ func adminGraphqlHandler(db *sql.DB, queries *sqlc.Queries, loaders *common.Batc
 	// Resolver is in the resolver.go file
 	h := handler.NewDefaultServer(gqladmingenerated.NewExecutableSchema(gqladmingenerated.Config{Resolvers: &resolver}))
 
+	directusSecret := config.Secrets.Directus
+	if directusSecret == "" {
+		log.L.Debug().Msg("No secret for Directus found in environment. Disabling endpoint")
+		return func(c *gin.Context) {
+
```
KillerX, line 68: IIRC this returns a 0 bytes 200 response. Maybe return a 404 or 401?

backend/items/collection/filter.go
```diff
@@ -0,0 +1,54 @@
...
+		case "desc":
+			orderByString += " DESC"
+		case "asc":
+			orderByString += " ASC"
+		}
+	}
+
+	queryString := "SELECT id FROM " + pq.QuoteIdentifier(collection) + " WHERE " + filterString + orderByString
+
+	rows, err := db.Query(queryString)
```
KillerX, line 49: I'm just wondering if we should add a debug log here for previews (if easy) so we could easily get see what SQL we actually executed if we need to debug something.

cms/src/endpoints/preview/index.ts
```diff
@@ -0,0 +1,67 @@
...
+                return
+            }
+            const filter = req.body.filter;
+            const collection = req.body.collection;
+
+            try {
+
+                const body = `query {
+                    preview {
+                        collection(collection: "${collection}", filter: ${JSON.stringify(JSON.stringify(filter))}) {
```
KillerX, line 35: Why do we need a double stringify? Is it basically an escape?
It would likely be best to write the queries with params here.

fredrikved: First stringify: Convert object to json. Second stringify: convert json object to string.

The filter isn't typed, so this is the way I send the filter to the api for deserializing. Know of a better way?

backend/items/collection/filter.go
fredrikved: Will log now, if preview is set in context (if that's sufficient)

KillerX (approved):

PR #95, opened 2022-08-16
Feat/query joins

fredrikved:
Not a very elegant way of solving it, but the relations in Directus is poorly documented, so this manual labour is necessary for now (can definitely revisit later)


KillerX (approved):
Not super pretty but I don't think stuff like this can ever be. I feel that the squirrel thingy gives me a bit more confidence that we can extend this if needed without having to pay too much attention to how changes in one part affect the other.

PR #101, opened 2022-08-22
Chore/permissions simplification

fredrikved:
Only 1 dataloader per "collection" should retrieve object data. The others only retrieve ids (relational, conversions, etc)

Permissions are retrieved separately as well, so we can check permissions independently


fredrikved:
Files and streams are only retrieved in 1 way, so they only have 1 loader with data which is also relational.

KillerX (approved):

PR #103, opened 2022-08-23
Feat/maintenance messages

fredrikved:
Implemented maintenance messages in API. 

Timestamp can be used to invalidate cache.


backend/graph/messages.go
```diff
@@ -0,0 +1,62 @@
...
+	return nil, false
+}
+
+var truncateTimeInterval time.Duration = 1
+
+func (r *Resolver) getMaintenanceMessages(ctx context.Context, timestamp *time.Time) ([]common.MaintenanceMessage, error) {
+	// Truncate time to avoid updating again within seconds
+	if timestamp != nil {
+		if timestamp.After(time.Now()) {
+			return nil, merry.New("invalid time (future)")
```
KillerX, line 34:
```suggestion
			timestamp = time.now()
```
We don't really "care" if people attempt to send in junk, so let's still return something reasonably sensible.

backend/graph/messages.go
```diff
@@ -0,0 +1,62 @@
...
+		truncated := timestamp.Truncate(time.Second * truncateTimeInterval)
+		timestamp = &truncated
+	}
+	if messages, ok := getMessagesFromCache(timestamp); ok {
+		return messages, nil
+	}
+
+	// Locking or waiting for another locked thread to complete
+	messageLock.Lock()
+	defer messageLock.Unlock()
```
KillerX, line 45: It feels like this will force all requests to be processed in a series. We don't really want that, do we?
I'm thinking something like this would make more sense:

```golang
// Global
lock := sync.RWMutex{}

// Local
lock.RLock()
	if messages, ok := getMessagesFromCache(timestamp); ok {
	        lock.RUnlock()
		return messages, nil
	}

// There is a small but non-zero chances that there are 2 threads in this section now
if lock.TryLock() {
messages, err := r.Queries.GetMaintenanceMessages(ctx)
	if err != nil {
		return nil, err
	}

	messageCache.Set("maintenance", timedCacheEntry{
		Cached: time.Now().Truncate(time.Second * truncateTimeInterval),
		Entry:  messages,
	}, cache.WithExpiration(time.Minute*10))
} else {
// Read from cache when it becomes available
lock.RLock()
	if messages, ok := getMessagesFromCache(timestamp); ok {
	        lock.RUnlock()
		return messages, nil
	} else {
 // ERROR
}
}
```

KillerX (dismissed):

PR #104, opened 2022-08-24
Don't require ShowID for new Season in the UI

KillerX:


fredrikved (changes requested):
No reason to update the schema file.

PR #105, opened 2022-08-25
Expose configuration

fredrikved:
Expose app configuration options

Probably need more configuration parameters.


backend/cmd/api/main.go
```diff
@@ -125,14 +126,13 @@ func main() {
 
 	collectionLoader := common.NewBatchLoader(queries.GetCollections)
 	// Set up urlSigner for CDN urls access
-	key, err := sign.LoadPEMPrivKeyFile(config.CDNConfig.AWSSigningKeyPath)
+	var key *rsa.PrivateKey
+	key, err = sign.LoadPEMPrivKeyFile(config.CDNConfig.AWSSigningKeyPath)
 	if err != nil {
-		log.L.Panic().Err(err).Msg("Unable to load PEM file")
-		panic(err)
+		log.L.Error().Err(err).Msg("Unable to load PEM file")
```
KillerX, line 132: Thanks!

queries/config.sql
```diff
@@ -0,0 +1,5 @@
+-- name: getAppConfig :one
+SELECT id, app_version FROM appconfig LIMIT 1;
+
+-- name: getGlobalConfig :one
+SELECT id, live_online, npaw_enabled FROM globalconfig LIMIT 1;
```
KillerX, line 5: This are both singletons, right?

KillerX (approved):

queries/config.sql
fredrikved: Yes.

PR #106, opened 2022-08-25
Realtime updates

fredrikved:
Post realtime updates to a firestore.

Infra not ready.

### DIFF
--- a/backend/cmd/jobs/env.go
+++ b/backend/cmd/jobs/env.go
@@ -35,6 +35,10 @@ type crowdinConfig struct {
 	ProjectIDs []int
 }
 
+type firebase struct {
+	ProjectID string
+}
+
 type envConfig struct {
 	AWS               awsConfig
 	Directus          directusConfig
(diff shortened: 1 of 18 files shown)


backend/cmd/jobs/env.go
```diff
@@ -35,6 +35,10 @@ type crowdinConfig struct {
 	ProjectIDs []int
 }
 
+type firebase struct {
+	ProjectID string
+}
```
KillerX, line 40: Can it implicitly use "current project id" if not set when running in GCP?

KillerX (commented):
Looks good imo, <del>maybe we should disable it if the PROJECT ID is not there?</del> Never mind, I see it.

backend/cmd/jobs/env.go
fredrikved: Probably, which variable is that stored in?

KillerX (approved):

fredrikved:
Just merging. Shouldn't affect anything yet anyway.

PR #118, opened 2022-09-08
Feat/auth0 access token

fredrikved:
Implement support for Auth0 access token (with retrieving data from Members API)


web/.env
```diff
@@ -1,5 +1,5 @@
-VITE_API_URL=http://localhost:8077/query
+VITE_API_URL=https://api.dev.brunstad.tv/query
 
-VITE_AUTH0_DOMAIN=btv-dev-test.eu.auth0.com
-VITE_AUTH0_CLIENT_ID=mOHwN9utx9OO5NSNcISB1F5sLtqWhlOA
-VITE_AUTH0_AUDIENCE=
\ No newline at end of file
+VITE_AUTH0_DOMAIN=bcc-sso-sandbox.eu.auth0.com
+VITE_AUTH0_CLIENT_ID=5HeUZ75GtalcU6BbxoWDs5u4YrGIZJ3s
+VITE_AUTH0_AUDIENCE=sandbox-api.bcc.no
```
KillerX: Maybe this should be in `sample.env`, as to not "configure" things automagically

KillerX (approved):

web/.env
fredrikved: .env is just the base. Any overrides should be added to .env.local, which vite reads as well as .env.production and .env.development.

I don't think there should be an issue with having it configured by default to the dev environment.

KillerX (dismissed):

KillerX (approved):

PR #122, opened 2022-09-12
Feat/improve initial setup

andreasgan:
See task #123


migrations/Makefile
```diff
@@ -0,0 +1,10 @@
+.PHONY: init install
+
+install:
+	npm install -g pg-diff-cli
+	go install github.com/pressly/goose/v3/cmd/goose@latest
+
```
KillerX, line 6: I'm in general not a fan of having this in makefiles. IMO the setup of the tools should not be the responsibility of the system, the system should just tell you what you are missing and suggest a copyable solution.

migrations/Makefile
```diff
@@ -0,0 +1,10 @@
...
+init: 
+	psql -h localhost -p 5432 -U btv < ./special/00-reset.sql
+	goose postgres "user=builder dbname=btv port=5432 host=localhost sslmode=disable" up
+	psql -h localhost -p 5432 -U btv btv < ./special/01-admin-user.sql
```
KillerX, line 10:
```suggestion
	psql -h localhost -p 5432 -U btv postgres < ./special/00-reset.sql
	goose postgres "user=btv dbname=btv port=5432 host=localhost sslmode=disable" up
	psql -h localhost -p 5432 -U btv btv < ./special/01-admin-user.sql
```

PR #126, opened 2022-09-13
Copy Video Player from old repository

fredrikved:
The package is published to NPM under the name btv-video-player.

Building it is also a bit weird. Will document.


fredrikved:
Made some of the endpoints non-nullable, as they should just return errors if not found when asking for them specifically.

KillerX (approved):

PR #128, opened 2022-09-13
Chore/documentation

fredrikved:


andreasgan:
@fredrikved I changed to `docker compose -d --wait" to wait for healthy. didnt help much because it wasnt truly ready immediately but yea, still nice to show that we tried lol.

But can you set up the project from scratch once to complete this fully? I still get stuck. I ran `make install` but it prompts for a db password for "btv".

andreasgan (changes requested):
see comment

andreasgan (dismissed):
I got it up and running now.
Haven't verified if the process is smooth fully from scratch but I think we're close.

cms/Makefile
```diff
@@ -11,15 +11,14 @@ build:
 
 s:
 	make build
-	docker-compose up -d
+	docker compose up -d --wait
```
KillerX, line 14: image
...

KillerX (approved):
I think we can just merge this and then work more in small iterations

KillerX (approved):

PR #135, opened 2022-09-14
To Master

KillerX:
Are we ok to go to STA and PROD or do we have some known issues/risks?


KillerX:
Are we ok to go to STA and PROD or do we have some known issues/risks?

fredrikved:
Idtokens from the app might break as I haven't tested with those, but the dev environment works as expected.

fredrikved (approved):

andreasgan:
idtokens should work before merging??

andreasgan:
@fredrikved

fredrikved:
Id-tokens work fine.

KillerX:
Setting this on hold until I can fix an ingest issue in PROD.

PR #137, opened 2022-09-15
Feat/improve pagination with permissions

fredrikved:
Paginate the collection item ids for permissions before loading the related data


KillerX (dismissed):
Just that "useless" error to look at, the rest seems ok.

KillerX (approved):

PR #142, opened 2022-09-20
Fake apply migration on diff

KillerX:
It asks before doing it.


KillerX:
> It asks before doing it.

You can also easily see if the migration number is correct:

image

fredrikved (approved):

PR #149, opened 2022-09-23
Chore/tracing

fredrikved:
Trace endpoint->database connections & compute requests


fredrikved:
Does this look right?

Ignore the changes to cms (using btv player instead of the old plugin for language selection and subtitles)

KillerX (approved):

PR #154, opened 2022-09-29
Lock crowdin sync so that multiple instances don't redo the sync

fredrikved:


backend/database/locks.go
```diff
@@ -0,0 +1,61 @@
+package database
```
KillerX, line 1: I generally like to leave (or add) credit to the original author, with link when I copy a whole function or file.

KillerX (approved):

PR #156, opened 2022-09-30
Rewrite the index.html to fill it with meta properties

fredrikved:
Added a simple go application which rewrites the urls on /episodes/:id, /seasons/:id, /shows/:id with public information retrieved from the /public endpoint.


backend/cmd/rewrite/api.go
```diff
@@ -0,0 +1,135 @@
...
+	if res.StatusCode < 200 || res.StatusCode > 299 {
+		log.Default().Print(merry.New("Error occurred when fetching", merry.WithHTTPCode(res.StatusCode), merry.WithMessage(string(resString))))
+		return nil
+	}
+
+	var r response[T]
+	err = json.Unmarshal(resString, &r)
+	if err != nil || r.Errors != nil {
+		return nil
+	}
```
KillerX, line 80: Log/Return?

backend/cmd/rewrite/api.go
```diff
@@ -0,0 +1,135 @@
...
+	Show struct {
+		Title       string
+		Description string
+		Image       *string
+	}
+}
+
+var apiEndpoint = os.Getenv("API_ENDPOINT")
+
+func get[T any](query string, variables map[string]any) *T {
```
KillerX, line 52: Any reason why we don't log most of the errors in this func?

backend/cmd/rewrite/api.go
```diff
@@ -0,0 +1,135 @@
...
+        image
+        number
+        season {
+            title
+            show {
+                title
+            }
+        }
+    }
+}
```
KillerX, line 99: Can we somehow easily have this in a file and "import" it into go at compile time? 
I just feel it's much nicer if the queries can be edited separately from the res of the code. 

This feels like it could be a super easy solution: https://pkg.go.dev/embed https://blog.jetbrains.com/go/2021/06/09/how-to-use-go-embed-in-go-1-16/

backend/cmd/rewrite/main.go
```diff
@@ -0,0 +1,195 @@
...
+	"log"
+	"net/http"
+)
+
+func (rw *rewriter) getDefaultHtml() *html.Node {
+	req, _ := http.NewRequest("GET", rw.webEndpoint, nil)
+	res, _ := http.DefaultClient.Do(req)
+
+	doc, _ := html.Parse(res.Body)
+	return doc
```
KillerX, line 16: I'm skeptical here as well about completely hiding errors. Should probably at least log them IMO

backend/cmd/rewrite/main.go
```diff
@@ -0,0 +1,195 @@
...
+}
+
+func addMetaProperty(node *html.Node, meta map[string]string) {
+	addTag(node, "meta", meta)
+}
+
+func (rw *rewriter) writeMeta(meta meta) string {
+	doc := rw.getDefaultHtml()
+	var f func(*html.Node)
+	f = func(n *html.Node) {
```
KillerX, line 58: Why not just have this as a normal function? Feels a bit excessive to define it dynamically

backend/cmd/rewrite/main.go
```diff
@@ -0,0 +1,195 @@
...
+	config := getEnvConfig()
+
+	rw := &rewriter{
+		apiEndpoint: config.APIEndpoint,
+		webEndpoint: config.WebEndpoint,
+	}
+
+	r := gin.Default()
+
+	r.GET("episodes/:id", func(ctx *gin.Context) {
```
KillerX, line 119: I would also prefer to extract the handler functions here. It makes it much easier to see what `main()` is doing, see the routes and the temptation to "randomly" capture vars from outside is much smaller :)

backend/cmd/rewrite/main.go
```diff
@@ -0,0 +1,195 @@
...
+			options.OGImage = *e.Image
+		}
+
+		h := rw.writeMeta(options)
+
+		ctx.Header("Content-Type", "text/html")
+		ctx.String(200, h)
+	})
+
+	_ = r.Run(":8076")
```
KillerX, line 194: Use PORT env variable, its a convention with CloudRun

backend/graph/api/collection-resolver.go
```diff
@@ -1,9 +1,9 @@
-package graph
+package graphapi
 
 import (
 	"context"
 	"github.com/bcc-code/brunstadtv/backend/common"
-	gqlmodel "github.com/bcc-code/brunstadtv/backend/graph/model"
+	gqlmodel2 "github.com/bcc-code/brunstadtv/backend/graph/api/model"
```
KillerX, line 6: Don't quite get why we have 2 here now?

backend/search/episodes.go
```diff
@@ -4,9 +4,12 @@ import (
 	"context"
 	"fmt"
 	"github.com/bcc-code/brunstadtv/backend/common"
+	"os"
 	"strconv"
 )
 
+var imageCDNDomain = os.Getenv("IMAGE_CDN_DOMAIN")
```
KillerX, line 11: Can we get this into `cmd`? The main reason why I packaged all of them together is to avoid sudden surprises about env vars you knew noting about being read deep in the code.

KillerX (changes requested):
One thing that was not quite clear to me is if this has any effect on the current API definition on the outside? I don't think it does but just need to be sure about that because the API is now in use by the app and we have to be careful meddling with it.

backend/graph/api/collection-resolver.go
fredrikved: Me neither. gqlgen has it's reasons I guess...
Will do some cleanup

KillerX: Ah, didn't realize this was codegen. Don't care in that case :)

backend/cmd/rewrite/main.go
fredrikved: Didn't realise recursiveness worked there.

fredrikved:
> One thing that was not quite clear to me is if this has any effect on the current API definition on the outside? I don't think it does but just need to be sure about that because the API is now in use by the app and we have to be careful meddling with it.

It shouldn't affect the current API at all. If for some reason it does, it is a bug, so fixable without any other issues.

backend/graph/api/collection-resolver.go
fredrikved: Do you think the new package names will create any issues?

They should never be mixed, so as long as you only import what you need, I think it makes sense (also if we add more endpoints later we won't run out of package names)

backend/search/episodes.go
fredrikved: No very clear way to do this, but config.ImageCDNDomain should override the use of this now (only used when retrieving from database)

backend/cmd/rewrite/main.go
fredrikved: Initial thought was that this was a simple application made to fail or return a default whenever it failed, but yes, there's no reason not to log actual errors.

backend/graph/api/collection-resolver.go
KillerX: No, I don't think it's an issue in general, just reacted to the non-descriptive name, but for genned code it's not an issue.

KillerX (approved):

PR #161, opened 2022-10-05
Filter list permissions in database request/dataloader

fredrikved:
Simple solution to filter retrieved ids in query, instead of in memory.


fredrikved:
Performance hit should be minimal, but I'll also do some tests.

backend/graph/api/collection-resolver.go
```diff
@@ -78,29 +78,77 @@ func collectionEntryResolver(ctx context.Context, loaders *common.BatchLoaders,
...
 }
 
 func collectionItemResolverFromCollection(ctx context.Context, r *Resolver, id string, first *int, offset *int) (*utils.PaginationResult[model.Item], error) {
 	int64ID, _ := strconv.ParseInt(id, 10, 32)
 
 	return collectionEntryResolver(ctx, r.Loaders, r.FilteredLoaders(ctx), int(int64ID), first, offset)
 }
+
+func sectionCollectionItemResolver(ctx context.Context, r *Resolver, id string, first *int, offset *int) (*utils.PaginationResult[*model.SectionItem], error) {
+	int64ID, _ := strconv.ParseInt(id, 10, 32)
```
KillerX, line 146: Not for this PR but maybe we should have a `util.AsInt()` by now 😅

KillerX (approved):

PR #162, opened 2022-10-06
Profiles and devices

fredrikved:
Device tokens should be filled by the client.


backend/graph/api/schema/schema.graphqls
```diff
@@ -83,4 +84,12 @@ type QueryRoot{
...
 
   config: Config!
+
+  profiles: [Profile!]!
+  profile: Profile!
+}
+
+type MutationRoot {
+  setDevice(device: DeviceInput!): Device
+  createProfile(name: String!): Profile
```
KillerX, line 94: I feel like I'm missing something here. To me looking at this as an app developer I have no idea what I'm supposed to use this for.

I would suggest:

* to have a automatic "default" profile for now and that creating of profiles is not exposed until we actually make use of that.
* to rename the "setDevice" to at least "setDevicePushToken" or something. It is very generic right now.

fredrikved: Profile is automatically created, and attached to the request if nothing else is specified.

Changed the endpoint for device tokens to accept a simple string instead.

KillerX (approved):

PR #166, opened 2022-10-11
Implement episode and season count on show

fredrikved:
One of two possible solutions.
This one: Retrieve episode & season count on request, filtered by permissions as well (cached by role)

Alternative: Total count of related episodes and seasons retrieved when retrieving show (globally cached)

I think this one is the best one for our use case, as long as we don't run into performance issues later.


KillerX (approved):
Looks good

PR #170, opened 2022-10-13
Expose build version

KillerX:
I think this is a reasonable way that does get in the way of any development and is easy to extend should we need to for some reason.

### DIFF
--- a/backend/cmd/api/main.go
+++ b/backend/cmd/api/main.go
@@ -28,6 +28,7 @@ import (
 	"github.com/bcc-code/brunstadtv/backend/sqlc"
 	"github.com/bcc-code/brunstadtv/backend/user"
 	"github.com/bcc-code/brunstadtv/backend/utils"
+	"github.com/bcc-code/brunstadtv/backend/version"
 	"github.com/bcc-code/mediabank-bridge/log"
 	"github.com/gin-contrib/cors"
 	"github.com/gin-gonic/gin"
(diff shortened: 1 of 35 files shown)


backend/version/handler.go
```diff
@@ -0,0 +1,70 @@
...
+
+		pwd = path.Join(pwd, "..")
+		pwd = path.Clean(pwd)
+		println(pwd)
+	}
+}
+
+// GinHandler returns a json string with the version data
+func GinHandler(ctx *gin.Context) {
+	ctx.Data(http.StatusOK, "application/json", versionJson)
```
fredrikved, line 53: ctx.JSON(200/http.StatusOK, obj) should work right

backend/cmd/api/main.go
```diff
@@ -329,6 +330,8 @@ func main() {
 
 	r.POST("/public", publicGraphqlHandler(loaders))
 
+	r.GET("/versionz", version.GinHandler)
```
fredrikved, line 333: Versionz?

fredrikved (dismissed):

backend/cmd/api/main.go
KillerX: Note the /versionz basically comes from https://stackoverflow.com/questions/43380939/where-does-the-convention-of-using-healthz-for-application-health-checks-come-f

fredrikved (approved):

PR #172, opened 2022-10-14
Implement link items

fredrikved:
Move from section type to collection item type


backend/graph/api/items.resolvers.go
```diff
@@ -134,7 +142,11 @@ func (r *showResolver) Image(ctx context.Context, obj *model.Show, style *model.
 	if style != nil && style.IsValid() {
 		s = style.String()
 	}
-	return e.Images.GetDefault(user.GetLanguagesFromCtx(ginCtx), s), nil
+	img := e.Images.GetDefault(user.GetLanguagesFromCtx(ginCtx), s)
+	if img == nil && e.Image.Valid {
+		img = &e.Image.String
+	}
+	return img, nil
```
KillerX, line 149: Would this bit of code (I see at least 3 repetitions) deserve to be packed into a named function? I think I know what it does but it would make it easier if it had a name directly on the struct. 

But if it's not "our" struct, don't bother.

backend/sqlc/section-queries.go
```diff
@@ -18,16 +17,12 @@ func mapToSections(items []getSectionsRow) []common.Section {
 		_ = json.Unmarshal(s.Description.RawMessage, &description)
 
 		t := "item"
-		if s.Type.Valid {
-			t = s.Type.ValueOrZero()
-		}
+		//if s.Type.Valid {
+		//	t = s.Type.ValueOrZero()
+		//}
```
KillerX, line 22: Should it be deleted completely?

KillerX (approved):

PR #173, opened 2022-10-17
Implement caching of profile

fredrikved:


backend/cmd/api/main.go
```diff
@@ -272,6 +273,13 @@ func main() {
 		return
 	}
 
+	rdb := redis.NewClient(&redis.Options{
+		Addr:     config.Redis.Address,
+		Password: config.Redis.Password,
+		Username: config.Redis.Username,
+		DB:       config.Redis.Database,
+	})
```
KillerX, line 281: We should do a ping here and fail loudly if it doesn't work (config).

backend/cmd/api/env.sample
```diff
@@ -29,3 +29,9 @@ AZ_SIGNING_KEY=<BASE 64 ENCODED SIMMETRYC KEY>
 
 # Other Secrets
 SERVICE_SECRET_DIRECTUS=<COMMON SECRET>
+
+# Redis
+REDIS_ADDRESS=
+REDIS_USERNAME=
+REDIS_PASSWORD=
```
KillerX, line 36: Is localhost default? If not you could add localhost to address so it works with compose out of the box.

KillerX (approved):
I'd just like to fail early if redis config is bad otherwise it looks good.

PR #174, opened 2022-10-17
Implement more push notifications capability

fredrikved:
It should be possible to push notifications to all registered devices now.


backend/push/client.go
```diff
@@ -23,36 +27,130 @@ func NewService(ctx context.Context, firebaseProjectID string) (*Service, error)
...
+		res, err := client.SendAll(ctx, r)
+		//TODO: Implement error handling
+		if err != nil {
+			return []error{err}
+		}
+		return lo.Map(lo.Filter(res.Responses, func(r *messaging.SendResponse, _ int) bool {
+			return r.Error != nil
+		}), func(r *messaging.SendResponse, _ int) error {
+			return r.Error
+		})
```
KillerX, line 87: Maybe a little comment here as to what is going on :) High energy lines of code.

backend/push/client.go
```diff
@@ -23,36 +27,130 @@ func NewService(ctx context.Context, firebaseProjectID string) (*Service, error)
...
+	})
+
+	for _, errs := range errors {
+		for _, err := range errs {
+			log.L.Error().Err(err).Msg("Error occurred when sending message")
+		}
```
KillerX, line 93:
```suggestion

			log.L.Error().Errs(errs).Msg("Error occurred when sending message")

```

KillerX (approved):
Some suggestions, but looks good.

PR #176, opened 2022-10-19
integration and lock test

fredrikved:
Initial integration test. Tests the redis lock.

Should probably move the things in the make file to a file or something.


KillerX (approved):

KillerX:
We can do that cleanup, but it's not a prio

PR #178, opened 2022-10-24
Episode Progress (continue watching ++)

fredrikved:
Implementation of storing episode progress data and retrieving it.


backend/cmd/api/main.go
```diff
@@ -94,13 +95,45 @@ func filteredLoaderFactory(db *sql.DB, queries *sqlc.Queries, collectionLoader *
...
+}
+
 // Defining the Graphql handler
 func graphqlHandler(db *sql.DB, queries *sqlc.Queries, loaders *common.BatchLoaders, searchService *search.Service, urlSigner *signing.Signer, config envConfig) gin.HandlerFunc {
 
 	resolver := graphapi.Resolver{
 		Queries:         queries,
 		Loaders:         loaders,
 		FilteredLoaders: filteredLoaderFactory(db, queries, loaders.CollectionLoader),
+		ProfileLoaders:  profileLoaderFactory(queries),
```
fredrikved, line 136: Thoughts on this?

Very simple queries will be performed here, but will still be towards the database without batching across users.
Not sure how we would batch across profiles and not overcomplicate it.

backend/graph/api/schema.resolvers.go
```diff
@@ -48,6 +48,44 @@ func (r *mutationRootResolver) SetDevicePushToken(ctx context.Context, token str
 	}, nil
 }
 
+// EpisodeProgress is the resolver for the episodeProgress field.
+func (r *mutationRootResolver) EpisodeProgress(ctx context.Context, id string, progress *int, duration *int) (*model.Episode, error) {
```
KillerX, line 52: Maybe the resolvers for mutations should include a verb, like `SetEpisodeProgress`?

backend/graph/api/schema.resolvers.go
```diff
@@ -48,6 +48,44 @@ func (r *mutationRootResolver) SetDevicePushToken(ctx context.Context, token str
...
+			EpisodeID: episodeID,
+			Progress:  *progress,
+			Duration:  dur,
+		}
+		err = q.SaveProgress(ctx, pr)
+		episodeProgress = &pr
+	}
+	pl := r.ProfileLoaders(ctx).ProgressLoader
+	pl.Clear(ctx, episodeID)
+	pl.Prime(ctx, episodeID, episodeProgress)
```
KillerX, line 85: Is this correct that we prime with nil in case of `progress==nil`?

backend/graph/api/schema/schema.graphqls
```diff
@@ -90,5 +90,5 @@ type QueryRoot{
 
 type MutationRoot {
   setDevicePushToken(token: String!): Device
-#  createProfile(name: String!): Profile
+  episodeProgress(id: ID!, progress: Int, duration: Int): Episode!
```
KillerX, line 93: Same as the resolver, would prefer a verb

backend/cmd/api/main.go
KillerX: Let's do it like this for now. We can't batch things that don't make sense.

KillerX (changes requested):
See comments ref resolver.

Another note: 
Would it make sense to "calculate" the "season progress" and "show progress" here?
I'm not thinking "how much of the show have you seen", but rather what "was the last episode you have watched in this show". It would simplify lookup later.

backend/graph/api/schema.resolvers.go
KillerX: Yes, redis would likely make more sense.

KillerX (approved):

PR #182, opened 2022-10-27
SQLite export

KillerX:
Consider this a preliminary review, not meant to be merged as-is.

This exports some of the key data that the user has access to, like shows, seasons, episodes, pages, collections, into a SQLite database.
The database is then copied to an S3 bucket (self cleaning), and a link that is valid for 1h is delivered back.

There are some general things to note:

* I don't particularly like the way the function itself is formatted but I don't want to make a major refactoring now for the sake of having a few less `if err != nil {}`
* The naming of the `serviceProvider` is just bad, but for the life of me I can't come up with something better right now. Possibly it is an indication that a refactoring is in order
* There is a big block of a comment on the version constant. Let me know if something does not make sense as this is basically our contract for our "SQLite api", and it would be quite bad if we broke things for many. 
* I am considering dropping goose since there is just one SQL file to run... But at the same time SQLc will split it into separate statements. And doing it in "raw" sql would require manual splitting on some level as well. Didn't test how much overhead in file size it adds. Doubt it has any impact on performance on other queries tho.
* Caching should be added regardless of the "need", just in case if errors and something going and requesting an export every 10 seconds. 
* Currently running `gqlgen` results in many changes in the generated files especially around pages and sections. Something that was forgotten? Maybe different version (v0.17.20 here)?


fredrikved (approved):
Would be interesting to see how big the file gets in production, if we need incremental updates or if a full sync is sufficient.

KillerX:
I'm merging this now so we can do a few tests.

PR #193, opened 2022-11-10
Initial implementation of related items

fredrikved:
This is an initial implementation. The collection id is hardcoded, as this will be changed when we do some computed things. 

Can probably change it now or make it configurable in admin.


fredrikved:
Now it's no longer hardcoded, but configured in App Config.

PR #198, opened 2022-11-15
Watched counter

fredrikved:


backend/graph/api/schema.resolvers.go
```diff
@@ -79,31 +79,45 @@ func (r *mutationRootResolver) SetEpisodeProgress(ctx context.Context, id string
...
+		episodeProgress, err = pl.Get(ctx, utils.AsInt(e.ID))
+		if err != nil {
+			return nil, err
 		}
-		var showID null.Int
-		if e.Season != nil {
-			s, err := r.QueryRoot().Season(ctx, e.Season.ID)
-			if err != nil {
-				return nil, err
+		if episodeProgress == nil {
```
fredrikved, line 90: should probably refactor due to the nesting here but I'm pretty sure we'll revisit it again

KillerX: I don't see a obvious way to do it either, so I guess its fine for now and we can extract into a method or something if really needed.

backend/graph/api/schema.resolvers.go
```diff
@@ -79,31 +79,45 @@ func (r *mutationRootResolver) SetEpisodeProgress(ctx context.Context, id string
...
 			}
-			showID.SetValid(int64(utils.AsInt(s.Show.ID)))
 		}
-		pr := common.Progress{
-			EpisodeID: episodeID,
-			ShowID:    showID,
-			Progress:  *progress,
-			Duration:  dur,
+
+		if float64(episodeProgress.Progress)/float64(episodeProgress.Duration) > 0.8 {
```
KillerX, line 112: I'd validate that `Duration > 0`, just in case so we don't blow up.

KillerX (approved):

PR #199, opened 2022-11-15
Implement continue watching collection/section

fredrikved:


backend/graph/api/collection-resolver.go
```diff
@@ -28,16 +30,55 @@ func preloadLoaders(ctx context.Context, loaders *common.BatchLoaders, entries [
...
+	if err != nil {
+		return nil, err
+	}
+
+	entries, err := collection.GetCollectionEntries(ctx, loaders, filteredLoaders, collectionId)
 	if err != nil {
 		return nil, err
 	}
 
+	if col.AdvancedType.Valid {
```
KillerX, line 57: Can we turn it around, to something like

```
if !col.AdvancedType.Valid {
     return ....
}
```

?

KillerX: Is `col.AdvancedType.String` of type `string` or `*string`.
If it's `string` then its `""` in case of `!Valid` so you can basically just drop the outside if, as ling as `""` doesn't have a meaning.

queries/progress.sql
```diff
@@ -1,5 +1,9 @@
 -- name: getProgressForProfile :many
-SELECT p.episode_id, p.show_id, p.progress, p.duration, COALESCE((p.progress::float / NULLIF(p.duration,0)) > 0.8, false)::bool AS watched
+SELECT p.episode_id,
+       p.show_id,
+       p.progress,
+       p.duration,
+       COALESCE((p.progress::float / NULLIF(p.duration, 0)) > 0.8, false)::bool AS watched
```
KillerX, line 6: How does SQL behave in division with 0?

KillerX (changes requested):
Some portential division by 0 stuff. I think its good to make sure they can't happen.

queries/progress.sql
fredrikved: It ends up as null.

But this will be changed in the other PR (with incremental watched instead of calculated on the fly)

backend/graph/api/collection-resolver.go
fredrikved: It doesn't return anything, it just changes the dataset, but i removed the if test as it's not necessary.

KillerX (approved):

PR #200, opened 2022-11-16
Feat/schedule notifications

fredrikved:
Implement scheduling notifications through Cloud Tasks.


backend/cmd/jobs/update-handler.go
```diff
@@ -0,0 +1,61 @@
...
+	queries   *sqlc.Queries
+	push      *push.Service
+	scheduler *scheduler.Service
+	locker    *redsync.Redsync
+}
+
+func (h *modelHandler) handleModelUpdate(ctx context.Context, collection string, key string) error {
+	switch collection {
+	case "notifications":
+		id := uuid.MustParse(key)
```
KillerX, line 26: Not sure if this is the best here. `must` by convention panics so if I manage to sneak something that's not UUID in, I can get instance to crash. Do that continuously and we have a DOS.

KillerX (dismissed):

KillerX (dismissed):

PR #203, opened 2022-11-17
Implement fetching redirect URLs with token

KillerX:


backend/graph/api/schema.resolvers.go
```diff
@@ -182,6 +186,59 @@ func (r *queryRootResolver) Export(ctx context.Context, groups []string) (*model
...
+
+	profile := user.GetProfileFromCtx(ginCtx)
+	if profile == nil {
+		return nil, merry.New(
+			"Not authorized",
+			merry.WithUserMessage("you are not authorized for this query"),
+		)
+	}
+
+	redir, err := r.Queries.GetRedirectByCode(ctx, id)
```
fredrikved, line 204: Do we want a cache here perhaps?

Put it through a dataloader and it should be cached, invalid ids as well.

migrations/00064_add_redirects.sql
```diff
@@ -0,0 +1,158 @@
...
+	"code" varchar(255) NOT NULL DEFAULT NULL::character varying ,
+	CONSTRAINT "redirects_pkey" PRIMARY KEY (id) ,
+	CONSTRAINT "redirects_user_updated_foreign" FOREIGN KEY (user_updated) REFERENCES directus_users(id) ,
+	CONSTRAINT "redirects_user_created_foreign" FOREIGN KEY (user_created) REFERENCES directus_users(id)
+);
+
+ALTER TABLE IF EXISTS "public"."redirects" OWNER TO builder;
+
+GRANT SELECT ON TABLE "public"."redirects" TO api; --WARN: Grant\Revoke privileges to a role can occure in a sql error during execution if role is missing to the target database!
+GRANT INSERT, INSERT, UPDATE, DELETE ON TABLE "public"."redirects" TO directus; --WARN: Grant\Revoke privileges to a role can occure in a sql error during execution if role is missing to the target database!
```
fredrikved, line 26: Duplicate insert grant? It probably doesn't fail but it's marked as an error here.

fredrikved (dismissed):

migrations/00064_add_redirects.sql
KillerX, line 26:
```suggestion
GRANT INSERT, UPDATE, DELETE ON TABLE "public"."redirects" TO directus; --WARN: Grant\Revoke privileges to a role can occure in a sql error during execution if role is missing to the target database!
```

backend/graph/api/schema.resolvers.go
KillerX: Yes, very good point. This is actually an action that is likely very spiky.

PR #204, opened 2022-11-18
Feat/settings

ThomasWongHKG:


web/src/layout/StackedLayout.vue
```diff
@@ -1,17 +1,19 @@
...
                 <component :key="$route.name" :is="Component" />
             </transition>
         </router-view>
-    </div>
-    <div class="text-red-500" v-if="errors && !loading">
-        <p v-for="(error, i) in errors">
-            {{ error.title }}
-            <span @click="removeError(i)" class="cursor-pointer">X</span>
-        </p>
+        <SettingContent :show="profileSettingIsShow" @closeDialog="profileToggle" />
```
fredrikved, line 10: If you figured out how to center it globally, I guess moving this back into where it's used is better. So in Navbar for example.

web/src/layout/StackedLayout.vue
```diff
@@ -1,17 +1,19 @@
 <template>
-    <Navbar v-if="!loading"></Navbar>
```
fredrikved, line 2: The HTML is not identical if you put it into a div, which is why I used a v-if test on each of the components.

I suggest keeping it as clean as possible (including the resulting HTML)

web/src/i18n/index.ts
```diff
@@ -4,7 +4,7 @@ import en from "./translations/en.json"
 
 import type { I18n, I18nOptions, Locale, Composer } from "vue-i18n"
 
-export const SUPPORT_LOCALES = ["en", "no"]
+export const SUPPORT_LOCALES = ["en", "no", "fr", "zh"]
```
fredrikved, line 7: Remove unsupported locales from this array, so you don't need to catch it later.

web/src/i18n/index.ts
```diff
@@ -32,7 +32,7 @@ export async function loadLocaleMessages(i18n: I18n, locale: Locale) {
 export default setup({
     legacy: false,
     locale: "en",
-    fallbackLocale: "en",
+    fallbackLocale: ['en'],
```
fredrikved, line 35: We're using double quotes usually everywhere. Do `pnpm prettier` to automatically fix most things.

web/src/layout/StackedLayout.vue
```diff
@@ -28,8 +30,19 @@ import Loader from "../components/Loader.vue"
...
+import { ref } from "vue"
+import SettingContent from '@/pages/setting/SettingContent.vue'
+
 provideClient(client)
 
+const profileSettingIsShow = ref(false);
+
+function profileToggle() {
+    console.log('profileToggle')
+    profileSettingIsShow.value = !profileSettingIsShow.value;
```
fredrikved, line 42: Not a fan of this syntax and variable naming. 

Use local variables as much as possible and handle toggles in template `@click="showSettings = !showSettings"`

web/src/pages/setting/SettingContent.vue
```diff
@@ -0,0 +1,107 @@
...
+  return result;
+}
+function getLocalSubSelected() {
+  if (localStorage.getItem('subSelected') == null) {
+    localStorage.setItem('subSelected', 'en')
+  }
+  var result: string = localStorage.getItem('subSelected') ?? 'en'
+  return result;
+}
+function getSelectedOption() {
```
fredrikved, line 45: This is not reactive.

Use refs and computeds where you need some reactive component and avoid the "function" syntax (use "() => {}" instead).

web/src/pages/setting/SettingContent.vue
```diff
@@ -0,0 +1,107 @@
...
+        leave-to="opacity-0">
+        <div class="fixed inset-0 bg-black bg-opacity-25" />
+      </TransitionChild>
+      <!-- Dialog Panel -->
+      <div class="flex min-h-full items-center justify-center p-4 text-center fixed inset-0 overflow-y-auto">
+        <TransitionChild as="template" enter-from="-translate-y-full" enter-to="translate-y-0"
+          enter="duration-300 ease-out" leave="duration-200 ease-in" leave-from="translate-y-0"
+          leave-to="-translate-y-full" class="min-w-[500px] ">
+          <DialogPanel
+            class="w-full max-w-md transform overflow-hidden rounded-2xl bg-blue-600 p-6 text-left align-middle shadow-xl transition-all">
```
fredrikved, line 75: Probably wrong color here. `bg-blue-600`. Look at figma for inspiration. If it's not stored in tailwind.config.cjs you can store a new one there.

web/src/components/settings/CxtLangPrefDragList.vue
```diff
@@ -0,0 +1,105 @@
+<template>
```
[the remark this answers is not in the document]
ThomasWongHKG: this is my testing file of the drag and drop component, it's for the language content list

web/src/layout/StackedLayout.vue
ThomasWongHKG: ok can I still make it like this
const settingsToggle = () => profileSettingIsShow.value = !profileSettingIsShow.value;

fredrikved: No thanks. 

Keep it in template where it needs to be. Shorten the variable names as well.

web/src/pages/setting/SettingContent.vue
ThomasWongHKG: the design of the dialog panel, I counld not find any similar design from the App figma, and I have raise this problem to Kim-David, So now waiting on his response on the design.

web/src/components/settings/CxtLangPrefDragList.vue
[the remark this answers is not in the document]
ThomasWongHKG: yeah, it was pushed accidential by inital create

web/src/pages/setting/SettingContent.vue
```diff
@@ -0,0 +1,98 @@
+<script lang="ts" setup>
```
fredrikved, line 1: Should probably rename this to SettingsModal or something like that. 

It's also not a page at the moment, so you should move it to components.

Pages is reserved for actual routes and subroutes.

web/src/pages/setting/SettingContent.vue
```diff
@@ -0,0 +1,98 @@
...
+  Tab,
+  TabPanels,
+  TabPanel,
+} from "@headlessui/vue"
+import LanguageSelector from "@/components/settings/LangSelector.vue"
+import { ref } from "vue"
+import { useI18n } from "vue-i18n"
+const { t } = useI18n()
+
+interface props {
```
fredrikved, line 27: Keep to the default type name convention (PascalCase).

In this case you probably don't need to define a separate interface, just put it directly into the type parameters on defineProps

web/src/pages/setting/SettingContent.vue
```diff
@@ -0,0 +1,98 @@
...
+  show: boolean
+}
+const props = defineProps<props>()
+
+const categories = ref({
+  audioLanguage: languages,
+  subtitleLanguage: languages,
+})
+const curretnCategory = ref(t("audioLanguage")) //inital value of the tab
```
fredrikved, line 36: Don't translate keys or logic components.

Minor typo as well (curretn)

web/src/pages/setting/SettingContent.vue
```diff
@@ -0,0 +1,98 @@
...
+
+const emits = defineEmits(["closeDialog"])
```
fredrikved, line 38: If you need to emit something, use the convention:

defineEmits<{
(event: "{key}", value: Type): void // or if you need something
}>()

If you don't assign it to a variable, the same functionality is available in template with $emit("{event}", value)

web/src/i18n/translations/en.json
```diff
@@ -42,5 +42,13 @@
...
         "episodes": "Episodes",
         "cancel": "Cancel"
+    },
+    "settings": {
+        "categories": {
+            "audioLanguage": "Audio Language",
+            "subtitleLanguage": "Subtitle Language"
+            
+        },
+        "profileSetting": "Profile Setting"
```
fredrikved, line 52: Use "Settings" and not the singular form of the word, and reevaluate the naming.

Don't think we need to refer to "Profile" visually for the user.

web/src/components/settings/LangSelector.vue
```diff
@@ -0,0 +1,51 @@
...
+  langlist: Language[] | LanguageWithIndex[]
+  selectedtab: string
+}
+const props = defineProps<Props>()
+
+const { selectedtab } = toRefs(props);
+
+const selectedOpt = ref(localStorage.getItem(selectedtab.value) ?? 'en')
+
+const updateHandler = (code: string) => {
```
fredrikved, line 17: Avoid having logic in components. Use props and emits from the parent. 

I don't think we should have code pointing to localStorage and other browser functions multiple places.

web/package.json
```diff
@@ -26,10 +26,13 @@
         "bccm-video-player": "^1.2.4",
         "graphql": "^16.6.0",
         "graphql-tag": "^2.12.6",
+        "sortablejs": "^1.14.0",
```
fredrikved, line 29: If you don't use a library, don't add them to the package.json (and include in the commit)

Same with vue-drag-drop and vue-draggable. They are packages which do the same job if I'm not mistaken. (and also unused)

fredrikved (changes requested):

PR #205, opened 2022-11-18
Feat/analytics

fredrikved:
Analytics on web implementation


backend/graph/api/schema/schema.graphqls
```diff
@@ -106,6 +106,8 @@ type QueryRoot{
   profiles: [Profile!]!
   profile: Profile!
 
+  analytics: Analytics!
```
KillerX, line 109: Would it make sense to have this under `{ me { analytics {...} } }`? Feels like user related data

KillerX (dismissed):

PR #215, opened 2022-12-06
Feat/studies

fredrikved:
TODO:
- Translations (Crowdin)
- Entrypoint (Pages/Sections)
- Permission filtering


migrations/00086_study_topics.sql
```diff
@@ -0,0 +1,2219 @@
+-- +goose Up
```
fredrikved, line 1: Just a bunch of table generation, shouldn't be any breaking table edits or anything in here. Some sorting, and some hiding of existing things, but nothing scary.

andreasgan (dismissed):
its too big for me to do a technical review but the schema LGTM, apart from the comment.

backend/graph/api/schema/studies.graphqls
```diff
@@ -0,0 +1,64 @@
...
+    first: Int!
+    total: Int!
+    items: [Task!]!
+}
+
+type AlternativesTask implements Task {
+    id: ID!
+    title: String!
+    completed: Boolean!
+    alternatives: [Alternative!]! @goField(forceResolver: true)
```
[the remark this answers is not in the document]
KillerX: Should that info be "published" or should that be an answer to a query?

backend/sqlc/study-queries.go
```diff
@@ -0,0 +1,165 @@
...
+	"gopkg.in/guregu/null.v4"
+)
+
+// GetTopics returns studies
+func (q *Queries) GetTopics(ctx context.Context, ids []uuid.UUID) ([]common.StudyTopic, error) {
+	topics, err := q.getTopics(ctx, ids)
+	if err != nil {
+		return nil, err
+	}
+	return lo.Map(topics, func(t getTopicsRow, _ int) common.StudyTopic {
```
KillerX, line 19: Would it make sense in this cases to use the parallel version of lo?

backend/utils/converters.go
```diff
@@ -0,0 +1,21 @@
...
+import (
+	"github.com/google/uuid"
+	"github.com/samber/lo"
+)
+
+// AsUuid parses string as Uuid, and returns it or an empty uuid.
+func AsUuid(s string) uuid.UUID {
+	uid, _ := uuid.Parse(s)
+	return uid
+}
```
KillerX, line 12: Should we have a warn log here, in case the UUID passed in is in fact invalid?

backend/graph/api/schema/schema.graphqls
```diff
@@ -127,4 +124,6 @@ type MutationRoot {
   setEpisodeProgress(id: ID!, progress: Int, duration: Int, context: EpisodeContext): Episode!
 
   sendSupportEmail(title: String!, content: String!, html: String!): Boolean!
+
+  completeTask(id: ID!, answer: TaskAnswer): Task!
```
KillerX, line 128: Should it maybe be 
```suggestion
  completeTask(id: ID!, answer: [TaskAnswer!]!): Task!
```
to support multi choice answers?

KillerX (commented):
A few questions but it looks ok in general.

backend/sqlc/study-queries.go
fredrikved: I don't think it would matter much. It's not waiting on any IO, just iterating through a list of items. 

Might be less efficient to split it into multiple threads.

backend/utils/converters.go
fredrikved: I don't think it'll help us with anything, as it just tells us if theres an invalid UUID _somewhere_ in the system. 

Maybe as a comment specify that it's only used when it is a uuid? At the moment I know for a fact that the only ids sent through the function are uuids, but in the odd case where it isn't, I'm fine with them being treated as empty uuids as well.

backend/graph/api/schema/schema.graphqls
fredrikved: Removed the answer field, as we aren't interested in the answers themselves at the moment, but I was thinking just a simple comma-separated answers whenever it was multiple select.

backend/utils/converters.go
KillerX: If you wrap the err with merry you could make it print the stack. That should give a clue/starting point.

backend/graph/api/schema/studies.graphqls
```diff
@@ -0,0 +1,60 @@
...
+    title: String!
+    completed: Boolean! @goField(forceResolver: true)
+    alternatives: [Alternative!]! @goField(forceResolver: true)
+}
+
+type Alternative {
+    id: ID!
+    title: String!
+    isCorrect: Boolean!
+    completed: Boolean!
```
andreasgan, line 51: `completed` er unødvendig her eller?

backend/graph/api/schema/studies.graphqls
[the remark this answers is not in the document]
andreasgan: we discussed and concluded on handling correct answers client-side, at least for this one. Less ux stuff to think about (loading states and error handling)

andreasgan (changes requested):

KillerX (approved):

andreasgan (approved):

PR #218, opened 2022-12-07
Feat/language setting

ThomasWongHKG:
This feature is to combine the old version, which is a simple content language dropdown list into including "Content Language", "Audio Language",  "Subtitle Language"


web/src/services/language.ts
```diff
@@ -79,3 +79,56 @@ export const setLanguage = async (l: string) => {
...
+        case 'Audio Language':
+            return currentAudioLanguage
+        case 'Subtitle Language':
+            return subtitleLanguage
+        default:
+            return;
+    }
+}
+
+export const setNestedLangMethod = (title: string, code: string) => {
```
fredrikved, line 123: You should probably not send "Title" here.

Use properly typed strings: `"content" | "audio" | "subtitle"`

web/src/pages/setting/playground.txt
```diff
@@ -0,0 +1,19 @@
+# default Scalar type {
```
fredrikved, line 1: Temporary files should not be included in a PR / commit

web/src/components/languages/LanguageSetting.vue
```diff
@@ -0,0 +1,111 @@
...
+import { current, languages, currentAudioLanguage, subtitleLanguage, getNestedLangComputed, setNestedLangMethod } from "@/services/language"
+import { onMounted, ref } from "vue"
+
+const selectedTitle = ref('')
+const isMenuActive = ref(true)
+
+defineProps<{ open: boolean }>()
+
+onMounted(() => {
+	var audLang = localStorage.getItem("audioLanguage") ?? "no"
```
fredrikved, line 12: Never use `var` in JS. 

Also, localstorage is references multiple places. I suggest moving this somewhere else.

fredrikved (changes requested):
Some test files still there. Also revisit the logic behind saving languages, as there's a bit of duplicate code.

PR #225, opened 2022-12-09
Add types to entries in collection

KillerX:


fredrikved (approved):
Wouldn't changes like these trigger a larger version increase? Removing a column?

Probably not important atm, but later I guess it would or?

KillerX:
Not before 1.0. As long as we have a 0 in front it's in a ALPHA/BETA stage.

PR #228, opened 2022-12-12
episode feedback

fredrikved:
Implement a way to submit feedback on episode


backend/graph/api/schema.resolvers.go
```diff
@@ -250,6 +252,76 @@ func (r *mutationRootResolver) UpdateTaskMessage(ctx context.Context, id string,
...
+	if err != nil {
+		return "", err
+	}
+	err = ratelimit.Endpoint(ctx, "episodes:messages:update", 100, false)
+	if err != nil {
+		return "", err
+	}
+	var metadata pqtype.NullRawMessage
+	if rating != nil {
+		metadata.RawMessage = []byte(fmt.Sprintf("{\"rating\": %d}", rating))
```
KillerX, line 307: I know it's annoying but can we do a struct and marshall it? :D
Can use `gin.H` (https://github.com/gin-gonic/gin/blob/v1.8.1/utils.go#L54) which is basically a dict. Something like:

```golang
		metadata.RawMessage = json.Marshall(gin.H{"rating", rating}}
```

backend/graph/api/schema.resolvers.go
```diff
@@ -250,6 +252,76 @@ func (r *mutationRootResolver) UpdateTaskMessage(ctx context.Context, id string,
...
+	if err != nil {
+		return "", err
+	}
+	var str string
+	if message != nil {
+		str = *message
+	}
+	var metadata pqtype.NullRawMessage
+	if rating != nil {
+		metadata.RawMessage = []byte(fmt.Sprintf("{\"rating\": %d}", rating))
```
KillerX, line 279: Same here.

KillerX (dismissed):

backend/graph/api/schema.resolvers.go
fredrikved: I don't disagree, and I cleaned up a bit more.

Map is fine for this right?

KillerX: Sounds good to me.

backend/graph/api/resolver.go
```diff
@@ -406,3 +411,89 @@ func getProfile(ctx context.Context) (*common.Profile, error) {
...
+	}
+	return episode, nil
+}
+
+func (r *Resolver) sendMessage(ctx context.Context, itemID uuid.UUID, message *string, metadata map[string]any) (string, error) {
+	err := ratelimit.Endpoint(ctx, "messages:send:"+itemID.String(), 2, false)
+	if err != nil {
+		return "", err
+	}
+	id, err := utils.GenerateRandomSecureString(32)
```
KillerX, line 450: This is the question for the function impl, but under what circumstances can this realistically fail?

KillerX (dismissed):

backend/graph/api/resolver.go
fredrikved: I'm pretty sure never. Could probably just ignore the error inside the function, as there's no way it will actually receive invalid input

fredrikved: In case you input a negative number it will return an error actually.

PR #229, opened 2022-12-12
Feat/translation improvement

fredrikved:
Improve translations


KillerX (approved):

KillerX:
Just make sure to validate it when it runs the first (few) time(s). I don't see any direct issues, but just to make sure the writing of works ok.

PR #231, opened 2022-12-16
Feat/achievements

fredrikved:
Implement achievements/badges in backend


backend/achievements/conditions.go
```diff
@@ -0,0 +1,94 @@
...
+	CollectionTasks   = "tasks"
+)
+
+type Action struct {
+	Collection string
+	Action     string
+}
+
+// CheckAchievements achieved since last check
+func CheckAchievements(ctx context.Context, queries *sqlc.Queries, loaders *common.BatchLoaders, action Action) error {
```
KillerX, line 25: Maybe the name should be something like "CheckForNewAchiviemnts"? Now it kinds sounds like you are checking the existing ones just reading the name.

backend/achievements/conditions.go
```diff
@@ -0,0 +1,94 @@
...
+		ProfileID:  p.ID,
+		Action:     action.Action,
+		Collection: action.Collection,
+		Amount:     int32(amount),
+	})
+	if err != nil {
+		return err
+	}
+	for _, aID := range ids {
+		err = queries.SetAchievementAchieved(ctx, sqlc.SetAchievementAchievedParams{
```
KillerX, line 64: Not sure where but it might make sense to log decisions when we decided to give an achievement, for debigging when the conditions are inevitably wrong :D

KillerX (approved):

backend/achievements/conditions.go
fredrikved: Would probably be enough to attach the condition ID?

It's probably easy to rerun achievements as well, and remove invalid assignments.

PR #235, opened 2023-01-03
Implement targeted notifications

fredrikved:
At the moment based only on usergroups, but queries can probably be built at a later stage.


fredrikved:
bcc-members and public roles not supported, as that requires an iteration of the entire userbase (not a problem, but not needed at the moment)

only respects the emails in the specific usergroups so we can test notifications

backend/members/lookup.go
```diff
@@ -3,9 +3,24 @@ package members
 import (
 	"context"
 	"fmt"
+	"github.com/cloudevents/sdk-go/v2/event/datacodec/json"
```
KillerX, line 6: Any specific reason for not using the builtin json package?

backend/members/lookup.go
```diff
@@ -3,9 +3,24 @@ package members
...
 
 // Lookup returns a member from the members api
 func (c *Client) Lookup(ctx context.Context, personID int) (*Member, error) {
 	return get[Member](ctx, c, fmt.Sprintf("persons/%d", personID))
 }
+
+func (c *Client) RetrieveByEmails(ctx context.Context, emails []string) (*[]Member, error) {
+	filter := struct {
+		Email map[string]any `json:"email"`
+	}{
```
KillerX, line 17: Personally I'm not a big fan of this kind of on the fly structs that are not empty. I'll say consider refactoring to a separate type (readability mostly).

KillerX (dismissed):
I didn't go super deep but seems logical.

backend/members/lookup.go
fredrikved: no. good spot. autoimport 🤷

PR #236, opened 2023-01-03
Feat/studytopics missing data

fredrikved:
Implement the missing study topic and lesson data. Images, availability in collections, etc.


backend/export/export.go
```diff
@@ -373,7 +373,8 @@ func exportCollections(ctx context.Context, q serviceProvider, liteQueries *sqle
 
 		liteEntries := lo.Map(entries, func(e collection.Entry, _ int) collectionEntry {
 			return collectionEntry{
-				ID:   e.ID,
+				// TODO: change this ID type to string
+				ID:   utils.AsInt(e.ID),
```
fredrikved, line 377: How's the compatibility for this now @KillerX ? Can we just update the structure here without bumping versions, etc?

KillerX: Until Africa tells us that they have released a prod version, just tag @oslabbie and let him know about the change. Do bump the patch version just so it's clear when this goes to prod.

KillerX (approved):

PR #245, opened 2023-01-13
Added competition mode to alternatives

KillerX:
Prevents you from submitting an answers to that task again

### DIFF
--- a/backend/cmd/api/main.go
+++ b/backend/cmd/api/main.go
@@ -137,7 +137,7 @@ func getLoadersForProfile(queries *sqlc.Queries, profileID uuid.UUID) *common.Pr
 			})
 		}, batchloaders.WithMemoryCache(time.Second*5)),
 		AchievementAchievedAtLoader:   batchloaders.New(profileQueries.GetAchievementsAchievedAt, batchloaders.WithMemoryCache(time.Second*5)),
-		GetSelectedAlternativesLoader: batchloaders.New(profileQueries.GetSelectedAlternatives, batchloaders.WithMemoryCache(time.Second*10)),
+		GetSelectedAlternativesLoader: batchloaders.New(profileQueries.GetSelectedAlternatives, batchloaders.WithMemoryCache(time.Second*5)),
 	}
 
 	profilesLoaderCache.Set(profileID, loaders, cache.WithExpiration(time.Minute*5))
--- a/backend/cmd/api/main.go
+++ b/backend/cmd/api/main.go
@@ -3,14 +3,15 @@ package main
 import (
 	"context"
 	"database/sql"
-	"github.com/99designs/gqlgen/graphql"
-	"github.com/ansel1/merry/v2"
-	"github.com/bcc-code/brunstadtv/backend/ratelimit"
-	"github.com/vektah/gqlparser/v2/gqlerror"
 	"net/http"
 	"sort"
 	"strings"
 	"time"
+
+	"github.com/99designs/gqlgen/graphql"
+	"github.com/ansel1/merry/v2"
+	"github.com/bcc-code/brunstadtv/backend/ratelimit"
+	"github.com/vektah/gqlparser/v2/gqlerror"
 
 	"github.com/bcc-code/brunstadtv/backend/analytics"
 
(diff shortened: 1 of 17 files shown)


backend/graph/api/schema.resolvers.go
```diff
@@ -173,36 +174,64 @@ func (r *mutationRootResolver) SendSupportEmail(ctx context.Context, title strin
...
+	if task.ComptetitionMode {
+		storedIds, err := r.Loaders.CompletedTasksLoader.Get(ctx, p.ID)
+		if err != nil {
+			return false, err
+		}
+
+		if lo.Contains(utils.PointerArrayToArray(storedIds), task.ID) {
+			return false, common.ErrTaskAlreadyCompleted
+		}
 	}
```
KillerX, line 197: Note: this is not entirely correct. There should be a "locked" flag that signifies the user has submitted the final answer.

backend/cmd/api/main.go
```diff
@@ -135,7 +136,8 @@ func getLoadersForProfile(queries *sqlc.Queries, profileID uuid.UUID) *common.Pr
 				Column2:   ids,
 			})
 		}, batchloaders.WithMemoryCache(time.Second*5)),
-		AchievementAchievedAtLoader: batchloaders.New(profileQueries.GetAchievementsAchievedAt, batchloaders.WithMemoryCache(time.Second*5)),
+		AchievementAchievedAtLoader:   batchloaders.New(profileQueries.GetAchievementsAchievedAt, batchloaders.WithMemoryCache(time.Second*5)),
+		GetSelectedAlternativesLoader: batchloaders.New(profileQueries.GetSelectedAlternatives, batchloaders.WithMemoryCache(time.Second*10)),
```
andreasgan, line 140: Well, 10 seconds is a bit much if I'm supposed to be stateless here.

andreasgan: But I'm considering to only load it once for a better UX on slower networks. But technically it would be best to reduce this down to the minimum.

backend/graph/api/schema.resolvers.go
```diff
@@ -173,36 +174,64 @@ func (r *mutationRootResolver) SendSupportEmail(ctx context.Context, title strin
...
-	if lo.Contains(utils.PointerArrayToArray(storedIds), task.ID) {
-		return false, common.ErrTaskAlreadyCompleted
+
+	selectedUUIDs := []uuid.UUID{}
```
andreasgan, line 199:
```suggestion
	var selectedUUIDs []uuid.UUID{}
```

I'm just thinking it can be confusing to have empty arrays in the DB?
Fredrik told me doing `append` with a nil slice like this works.

andreasgan (changes requested):

backend/common/studies.go
```diff
@@ -57,19 +57,20 @@ func (i Lesson) GetKey() uuid.UUID {
...
+	LessonID         uuid.UUID
+	Title            LocaleString
+	Type             StudyTaskType
+	QuestionType     QuestionTaskType
+	ImageType        ImageTaskType
+	LinkID           null.Int
+	EpisodeID        null.Int
+	Images           LocaleMap[string]
+	MultiSelect      null.Bool
+	ComptetitionMode bool
```
andreasgan, line 70:
```suggestion
	CompetitionMode bool
```

andreasgan (changes requested):

backend/cmd/api/main.go
KillerX: I can see if I can easily evict that key from cache when you submit an answer.

backend/graph/api/schema.resolvers.go
KillerX: I'll take a look, but yeah it would make sense.

backend/cmd/api/main.go
KillerX: I mean 10s is basically since you loaded the answer so you would have to load answer select, submit, go back. Not impossible but somewhat reasonable I feel, I'll reduce it to 3 seconds for the PR, and then look into better options after. OK?

andreasgan: I was thinking more if I select something, go to the next task and then back to edit.. But it's a better UX to just remember the state so that's what I'm going to do.

KillerX: It's easy to clear cache, doing that on submission now.

andreasgan (approved):

PR #246, opened 2023-01-13
feat: order tasks by sort

andreasgan:
Can I merge this? @KillerX


queries/studies.sql
```diff
@@ -75,7 +75,8 @@ FROM tasks t
          LEFT JOIN ts ON ts.tasks_id = t.id
          LEFT JOIN images ON images.task_id = t.id
 WHERE t.status = 'published'
-  AND t.id = ANY ($1::uuid[]);
+  AND t.id = ANY ($1::uuid[])
+ORDER BY t.sort;
```
KillerX, line 79: Add `ASC` to be explicit.

andreasgan: we don't do that anywhere else tho

PR #258, opened 2023-01-26
Wrong completion count

KillerX:


backend/user/middleware.go
```diff
@@ -159,6 +159,7 @@ func NewUserMiddleware(queries *sqlc.Queries, membersClient *members.Client) fun
 			span.AddEvent("User failed to load")
 			userCache.Set(userID, u, cache.WithExpiration(1*time.Minute))
 			ctx.Set(CtxUser, u)
+			return
```
KillerX, line 162: This should have been there all along. We bail out if there is an error getting person data, not related to this pr but it slipped in.

queries/studies.sql
```diff
@@ -161,7 +162,7 @@ WITH total AS (SELECT t.lesson_id,
 SELECT total.lesson_id as id, p.id as parent_id
 FROM users.profiles p
          JOIN completed ON completed.profile_id = p.id
-         JOIN total ON total.lesson_id = completed.lesson_id
+         JOIN total ON total.lesson_id >= completed.lesson_id -- In case somethig has been archived later
```
fredrikved, line 165: >= doesn't make sense here on the join or?
Wouldn't there be a lot of rows?

KillerX: Loooool I put this in a totally wrong place :D. Gimme a moment.

KillerX: Fixed.

fredrikved (approved):

PR #260, opened 2023-01-27
Stats in admin

KillerX:
Added a new endpoint to Admin GQL to deliver some stats.


backend/graph/admin/schema.resolvers.go
```diff
@@ -63,11 +65,53 @@ func (r *queryRootResolver) Preview(ctx context.Context) (*model.Preview, error)
...
+		AgeGroups: ageGroups,
+		LessonID:  utils.AsUuid(lessonID),
+		MinSize:   int32(minSize),
+		MaxSize:   int32(maxSize),
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	out := []*model.ProgressByOrg{}
```
fredrikved, line 95: lo.Map for cleanness sake?

backend/members/lookup.go
```diff
@@ -27,3 +30,134 @@ func (c *Client) RetrieveByEmails(ctx context.Context, emails []string) (*[]Memb
...
+
+	encoded, _ := json.Marshal(filter)
+
+	return get[[]Member](ctx, c, fmt.Sprintf("persons?limit=999&fields=personID,age,affiliations.*&filter=%s", encoded))
+}
+
+// CountMembersByAge counts active members who are min <= age <= max
+func (c *Client) CountMembersByAge(ctx context.Context, min, max int) (int, error) {
+	/// TOOO: This is currenty only used by `statistician` and needs to be cleaned up, possibly limited?
+	filter := gin.H{
```
fredrikved, line 56: Instead of using a package alias for map[string]any, mb create a local/internal shorthand for it? Or does it have some tooling you use? Kind of weird to reference external packages for simple things like these, but that's my thoughts.

backend/members/lookup.go
```diff
@@ -27,3 +30,134 @@ func (c *Client) RetrieveByEmails(ctx context.Context, emails []string) (*[]Memb
...
+		}
+
+		// Are we done?
+		if len(*ms) == 0 {
+			return out, nil
+		}
+
+		// Can this easily be parallelized?
+		// Is the += thread safe or do we have to make it better
+		// Will it be faster? Do we care?
```
fredrikved, line 121: If I'm not mistaken, this is an admin induced function, and only for statistical reasons, so I don't think we will benefit much from splitting the thread, even if it reduces the overall execution time. Better to keep the computational complexity to a minimum I think.

PR #261, opened 2023-01-30
Expose VOD Player for offline use of BTV

oslabbie:


KillerX:
@oslabbie this is needed so you can interact with the player right?

oslabbie:
Yes that is correct

KillerX:
@fredrikved  are you able to connect with @oslabbie and figure out what we can do to help since you mentioned that this PR would likely have negative consequences.

fredrikved (approved):

PR #263, opened 2023-01-31
Allow editing of schedule

fredrikved:
Open up for editing a scheduled task.

### DIFF
--- a/backend/scheduler/client.go
+++ b/backend/scheduler/client.go
@@ -5,8 +5,11 @@ import (
 	"cloud.google.com/go/cloudtasks/apiv2/cloudtaskspb"
 	"context"
 	"encoding/json"
+	"github.com/bcc-code/brunstadtv/backend/utils"
 	"github.com/gin-gonic/gin"
+	"google.golang.org/api/iterator"
 	"google.golang.org/protobuf/types/known/timestamppb"
+	"strings"
 	"time"
 )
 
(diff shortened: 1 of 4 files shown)


backend/scheduler/client.go
```diff
@@ -39,12 +42,38 @@ type QueuedItem struct {
...
+		Parent: s.queueID,
+	})
+
 	taskName := s.queueID + "/tasks/" + collection + "-" + id
 
+	for {
+		resp, err := it.Next()
+		if err == iterator.Done {
+			break
+		}
```
KillerX, line 55:
```suggestion
	for resp, err := it.Next(); err != iterator.Done {
```

I *think* this is valid go code. 
Don't have to use it if you don't like it. I was just looking at the pattern for getting SQL rows, but its iterator operates differently:

```
for rows.Next() {
	err := rows.Scan(&id, &name)
	if err != nil {
		log.Fatal(err)
	}
	log.Println(id, name)
}
```

backend/utils/random.go
```diff
@@ -6,16 +6,13 @@ import (
 )
 
 // GenerateRandomSecureString generates a random secure string
-func GenerateRandomSecureString(length int) (string, error) {
+func GenerateRandomSecureString(length int) string {
```
KillerX, line 9: FYI: https://github.com/samber/lo#randomstring since we are already using `lo` in lots of places

KillerX (approved):
LGTM: Just some notes, but nothing important

PR #264, opened 2023-01-31
Chore/refactoring

fredrikved:
Refactor references to loaders, so that it's a more streamlined approach everywhere.


fredrikved:
I believe we might have some issues with janitors running away, so going to try to fix that as well in this pr.

backend/cmd/api/loaders.go
```diff
@@ -0,0 +1,173 @@
...
+// I doubt instantiating this 10000 times will be slow,
+// and we'll probably have fewer issues with runaway janitors.
+var profileLoaders = sync.Map{}
+
+func getLoadersForProfile(queries *sqlc.Queries, profileID uuid.UUID) *common.ProfileLoaders {
+	if ls, ok := profileLoaders.Load(profileID); ok {
+		return ls.(*common.ProfileLoaders)
+	}
+
+	ctx := context.TODO()
```
fredrikved, line 72: Passing a cancelable context into the loaders allows us to control when they terminate later. 

For the moment, I'm letting them live, and later I will cancel the contexts when removing the key from the map.

PR #266, opened 2023-01-31
add availability tests

starkris51:


tests/e2e/lib/apiResponseFromEpisode.ts
```diff
@@ -0,0 +1,75 @@
...
+        })
+
+    return apiResponse
+}
+
+export async function getApiResponsefromEpisodeWithAvailability(title: string, available_from?: Date, available_to?: Date) {
+    const episode = await client.items("episodes").createOne({
+        status: "published",
+        available_from: available_from == null ? null : available_from.toISOString(),
+        available_to: available_to == null ? null : available_to.toISOString(),
```
fredrikved, line 43: Even in tests, we should probably avoid `==` at all costs, as they are misleading and can create problems down the line. Always use `===` or alternative methods to get the result you want.

In this case I would either write `available_from ? available_from.toISOString() : null` or a shorthand for it: `available_from?.toISOString() ?? null`

fredrikved (approved):

PR #270, opened 2023-02-06
Chore/redis refactor

fredrikved:
Start refactoring redis. 

Changed the packages or at least references to packages, and use a more official approach to locking.

### DIFF
--- a/backend/remotecache/cache.go
+++ b/backend/remotecache/cache.go
@@ -56,7 +56,9 @@ func GetOrCreate[T any](ctx context.Context, rc *Client, key string, valueFactor
 	if err != Nil {
 		return value, err
 	}
-	options := &Options{}
+	options := &Options{
+		expiry: time.Minute,
+	}
 	lock, err := Lock(ctx, rc.Locker(), key)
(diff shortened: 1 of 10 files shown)


backend/remotecache/cache.go
```diff
@@ -0,0 +1,68 @@
...
+		return value, err
+	}
+	options := &Options{}
+	lock, err := Lock(ctx, rc.Locker(), key)
+	defer Release(ctx, lock)
+	value, err = valueFactory(options)
+	if err != nil {
+		return value, err
+	}
+	err = Set(ctx, rc.Client(), key, value, options.expiry)
```
KillerX, line 66: What will the default be if `options.expiry` is not changed? 0? Do we then even bother to write it to the cache?

KillerX (approved):

backend/remotecache/cache.go
fredrikved: 0 yes. Can't not include it, but 0 means infinite TTL. 

Want a default? F.ex 5 minutes?

PR #273, opened 2023-02-07
Implement backup and remote caching of members response

fredrikved:


backend/cmd/api/main.go
```diff
@@ -172,7 +173,20 @@ func main() {
 	queries := sqlc.New(db)
 	queries.SetImageCDNDomain(config.CDNConfig.ImageCDNDomain)
 	authClient := auth0.New(config.Auth0)
-	membersClient := members.New(config.Members, authClient)
+
+	cb := gobreaker.NewCircuitBreaker(gobreaker.Settings{
+		Name:    "Members",
+		Timeout: time.Second * 2,
+		ReadyToTrip: func(counts gobreaker.Counts) bool {
+			failureRatio := float64(counts.TotalFailures) / float64(counts.Requests)
```
KillerX, line 181: Is counts.Requests guaranteed to to be `!= 0`?

backend/user/middleware.go
```diff
@@ -135,78 +137,114 @@ func NewUserMiddleware(queries *sqlc.Queries, memberLoader *loaders.Loader[int,
...
-					// Note: Maps are not iterated in a sorted order, so we have to find the lowed applicable
-					if u.Age >= minAge && minAge > ageGroupMin {
-						u.AgeGroup = group
-						ageGroupMin = minAge
+				roles = append(roles, userRoles...)
+			}
+
+			u.Roles = roles
+			u.Email = email
+			u.DisplayName = member.DisplayName
```
KillerX, line 193: Not for this PR but this does not contain what (I would have) expected to be a "first + last" style string.
For example everyone in Mallorca only has the First Name in there. Something to look at down the line.

backend/user/middleware.go
```diff
@@ -135,78 +137,114 @@ func NewUserMiddleware(queries *sqlc.Queries, memberLoader *loaders.Loader[int,
...
+
+			// Set AgeGroup and avoid passing identifying information through the application
+			if member.BirthDate != "" {
```
KillerX, line 196: Are you aware that members provides a field called `qge` so we could drop the whole parsing part?

backend/user/middleware.go
```diff
@@ -135,78 +137,114 @@ func NewUserMiddleware(queries *sqlc.Queries, memberLoader *loaders.Loader[int,
...
-		}), func(i members.Affiliation, _ int) int {
-			return i.OrgID
-		})
+			u.ChurchIDs = lo.Map(lo.Filter(member.Affiliations, func(i members.Affiliation, _ int) bool {
+				return i.Active && i.OrgType == "Church"
+			}), func(i members.Affiliation, _ int) int {
+				return i.OrgID
+			})
+
+			err = queries.InsertUser(ctx, sqlc.InsertUserParams{
```
KillerX, line 219: Maybe rename to `UpsertUser`?
(Yes, I was the lazy one :) )

migrations/00127_users_users.sql
```diff
@@ -0,0 +1,71 @@
...
+
+--- BEGIN CREATE TABLE "users"."users" ---
+
+CREATE TABLE IF NOT EXISTS "users"."users" (
+	"id" varchar NOT NULL  ,
+	"email" varchar NOT NULL  ,
+	"display_name" varchar NULL  ,
+	"age" int4 NULL  ,
+	"church_ids" _int4 NULL  ,
+	"active_bcc" bool NULL  ,
```
KillerX, line 15: Any special reason for this to be nullable?

KillerX (commented):
Some small comments but seems good otherwise.

migrations/00127_users_users.sql
fredrikved: Was for debug reasons, but no. Will fix the small issues

backend/user/middleware.go
fredrikved: I don't think this field is important for us, but it is what we should use for displaying info or for any interactions.

backend/cmd/api/main.go
fredrikved: Think I'm going to keep it as simple as possible for the initial implementation.

KillerX (dismissed):

PR #274, opened 2023-02-07
Make Nathan Lux not have an active BCC Membership

KillerX:


KillerX:
@fredrikved  unfortunately this is required for now. Ill merge and test a bit.

PR #276, opened 2023-02-08
Feat/k6 load tests

fredrikved:
Initial implementation of load tests.

Look at the changes in api/main.go primarily. Need to be able to write tests towards a real target (staging), I think.

### DIFF
--- a/backend/graph/api/resolver.go
+++ b/backend/graph/api/resolver.go
@@ -472,12 +472,10 @@ func (r *Resolver) sendMessage(ctx context.Context, itemID uuid.UUID, message *s
 		insertParams.AgeGroup = usr.AgeGroup
 	}
 
-	if !user.IsImpersonating(gc) {
-		err = r.Queries.SetMessage(ctx, insertParams)
-		if err != nil {
-			log.L.Error().Err(err).Msg("Failed to save string to database")
-			return "", merry.New("Failed to generate unique ID")
-		}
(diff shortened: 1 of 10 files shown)


fredrikved:
Added IsImpersonating to avoid writing to db where that might be the case.

backend/graph/api/resolver.go
```diff
@@ -480,10 +472,12 @@ func (r *Resolver) sendMessage(ctx context.Context, itemID uuid.UUID, message *s
...
-	err = r.Queries.SetMessage(ctx, insertParams)
-	if err != nil {
-		log.L.Error().Err(err).Msg("Failed to save string to database")
-		return "", merry.New("Failed to generate unique ID")
+	if !user.IsImpersonating(gc) {
+		err = r.Queries.SetMessage(ctx, insertParams)
+		if err != nil {
+			log.L.Error().Err(err).Msg("Failed to save string to database")
+			return "", merry.New("Failed to generate unique ID")
+		}
```
KillerX, line 480: Why are we not doing it here? I feel that this mutations are an important part of the load test...

fredrikved: It's important yes, so all the "impersonating" things can probably be changed or ignored, but I don't want to write unnecessarily or by mistake to the database until we have a proper staging or testing environment.

I don't think I will be doing any mutations until I've made the fetch requests, and when we need to test writes, we can easily just remove these checks.

KillerX (approved):

PR #281, opened 2023-02-13
Materialized view for the filter dataset

fredrikved:
Change to a materialized view for the filter dataset.

What's your thoughts on having updates triggering a refresh internally? Would it be a performance issue?


KillerX:
It of course depends how often the updates happen. Ideally you would have some sort of cooldown period (f.ex. max one update per x seconds/minutes and you just queue one up during that time if nothing is waiting yet, otherwise you just ignore it. 
We can also explore the (not sure if there is a formal name) self updating table that gets recalculated in triggers but not completely rebuilt, only rows that are affected by the update. Tends to be more complex but more performant and "real time".

fredrikved:
That's probably something we can explore later. For now though, I want to see if this can fix some of the slow queries.

PR #282, opened 2023-02-14
Allow sending notifications to only members

fredrikved:
Notifications can be filtered by the roles: bcc-members or registered. Registered will target everyone regardless of active membership and bcc-members will target every active member.


fredrikved:
Seems I committed originally to develop. 🙃 

Look at the previous commit as well if you can.

fredrikved:
Never mind. Reverted the commit and added the changes here.

KillerX (approved):
As far as I understand it seems sane.

PR #287, opened 2023-02-22
Expose unlisted status

fredrikved:


backend/graph/api/model/episode.go
```diff
@@ -64,9 +64,15 @@ func EpisodeFrom(ctx context.Context, e *common.Episode) *Episode {
 		episodeType = EpisodeTypeStandalone
 	}
 
+	status := StatusPublished
+	if e.Unlisted() {
+		status = StatusUnlisted
+	}
```
andreasgan, line 70: This was very confusing, because there is a common.Status and the model.Status. They have the same name for the type and the constants.

At first, I was thinking: "Why cant you just use the status directly"? And what about the other values?
I understand what's going on, but it's confusing.

It would help with a mapping function like: `status := statusFrom(e.Status)`

andreasgan (changes requested):

andreasgan:
otherwise lgtm

backend/graph/api/model/episode.go
fredrikved: The GQL API is just one of the interfaces that we create from the common models, and in some cases the bg-worker might use the data there differently than what the GQL API does.

The other values are irrelevant for the API at the moment, so I see no reason exposing them. Just fallback to published if they do not serve a specific purpose.

backend/graph/api/model/status.go
```diff
@@ -0,0 +1,12 @@
+package model
+
+type canBeUnlisted interface {
+	Unlisted() bool
```
andreasgan, line 4: Isnt this relying on an implementation detail? this mapping function is only to map statuses from graphql.
No need to make it too generic, it hurts readability.

andreasgan: @fredrikved

fredrikved: It doesn't matter too much. Super easy to change. I find this better than referencing common.Status. 

If we need more statuses later, we just reimplement this in another way.

andreasgan: Yeah but in general if you do this everywhere it can get messy. something to consider

PR #289, opened 2023-02-27
Surveys

fredrikved:
Implement surveys and a solution for app to listen to changes


KillerX:
Do we have a background for this. It looks ok but it's hard to understand why it's there :D

fredrikved:
Actually something I hadn't considered yet, so will update this tomorrow.

KillerX:
Also update the migration numbers, I just had to add one

KillerX (approved):

PR #292, opened 2023-03-03
Feat/user collections

fredrikved:
Implement user collections / My List.


backend/graph/api/schema/mutations.graphqls
```diff
@@ -21,8 +21,17 @@ type MutationRoot {
 
     answerSurveyQuestion(id: UUID!, answer: String!): AnswerSurveyQuestionResult!
     updateSurveyQuestionAnswer(key: String!, answer: String!): AnswerSurveyQuestionResult!
+
+    addEpisodeToMyList(episodeId: ID!): AddToCollectionResult!
+    addShowToMyList(showId: ID!): AddToCollectionResult!
```
KillerX, line 26: Would make more sense to just return the `UserCollection`? 
Not sure how it will be used, maybe @andreasgan has some more opinion here?

backend/graph/api/schema/schema.graphqls
```diff
@@ -120,6 +120,10 @@ type QueryRoot{
 
   me: User!
 
+  myList: UserCollection!
```
KillerX, line 123: Do we only have 1 list? Or is this a special/default one?
Specific per app & profile?

migrations/00151_user_collections.sql
```diff
@@ -0,0 +1,122 @@
...
+CREATE TABLE IF NOT EXISTS "users"."collections" (
+                                                     "id" uuid NOT NULL  ,
+                                                     "profile_id" uuid NOT NULL  ,
+                                                     "updated_at" timestamp NOT NULL  ,
+                                                     "created_at" timestamp NOT NULL  ,
+                                                     "metadata" json NULL  ,
+                                                     "title" text NOT NULL  ,
+                                                     CONSTRAINT "collections_pk" PRIMARY KEY (id) ,
+                                                     CONSTRAINT "collections_profiles_id_fk" FOREIGN KEY (profile_id) REFERENCES users.profiles(id) ON UPDATE CASCADE ON DELETE CASCADE
+);
```
KillerX, line 18: Do we need to have lists per APP?

KillerX (commented):
Some general comments, about the API that I think we need to address before merging this.

backend/graph/api/schema/schema.graphqls
fredrikved: At the moment it's only 1 list. 

I wanted the same structure though in case we want users to be able to create multiple lists down the line.

migrations/00151_user_collections.sql
fredrikved: Good idea per app. Will add.

backend/graph/api/schema/mutations.graphqls
fredrikved: It depends on typing. Not sure yet if we want to return more metadata or if the result should be different. Having a type for it only slightly increases clutter, but simplifies additions in the future.

backend/graph/api/schema/schema.graphqls
fredrikved: Actually filtered by a parameter in the metadata field, so only 1 per profile & app.

andreasgan:
Isn't it annoying and unsafe to use a json meta field for this? Can't it just be a column?

fredrikved:
> Isn't it annoying and unsafe to use a json `metadata` column for this? Can't it just be a column?

Is there any difference that would affect performance as it is now?

I believe MyList functionality is a temporary solution and would be replaced later down the line, so having a column for this and that would clutter the database in the end.

andreasgan:
> > Isn't it annoying and unsafe to use a json `metadata` column for this? Can't it just be a column?
> 
> Is there any difference that would affect performance as it is now?
> 
> I believe MyList functionality is a temporary solution and would be replaced later down the line, so having a column for this and that would clutter the database in the end.

If you're going to migrate away from it, you're gonna want to remove the "mylist" data anyway, so what's the difference between cleaning up the json column and deleting the column?
Having it structured in a column makes migrating to a different solution even easier?
A json column is just as much clutter, but you lose data integrity and increase complexity (e.g. marshalling etc)?
Also, "nothing is as permanent as a temporary solution".

PR #294, opened 2023-03-08
implement prompts instead of surveys on root

fredrikved:


backend/graph/api/schema/surveys.graphqls
```diff
@@ -31,3 +29,14 @@ type SurveyRatingQuestion implements SurveyQuestion {
...
 }
+
+interface Prompt {
+    id: UUID!
+    title: String!
+}
+
+type SurveyPrompt implements Prompt {
+    id: UUID!
+    title: String!
```
andreasgan, line 40: missing the from/to fields

andreasgan (changes requested):
did you implement a firestore field for this?

andreasgan (approved):

PR #295, opened 2023-03-13
Check that email is verified, or filter roles

fredrikved:


backend/user/middleware.go
```diff
@@ -31,6 +31,9 @@ const (
 	RoleNonBCCMember = "non-bcc-members"
 )
 
+// DefaultRoles are default roles in the system.
+var DefaultRoles = []string{RolePublic, RoleRegistered, RoleBCCMember, RoleNonBCCMember}
```
andreasgan, line 35: const?

backend/user/middleware.go
```diff
@@ -228,7 +234,13 @@ func NewUserMiddleware(queries *sqlc.Queries, remoteCache *remotecache.Client, l
 				roles = append(roles, userRoles...)
 			}
 
-			u.Roles = roles
+			if !u.EmailVerified {
```
andreasgan, line 237: Any reason not to skip the "GetRolesForEmail" step instead of doing this filtering after?

andreasgan (changes requested):

backend/user/middleware.go
fredrikved: arrays cant be consts, as they are in nature dynamic

andreasgan: oh ok.
but you can make fixed size arrays with [...]
https://stackoverflow.com/questions/13137463/declare-a-constant-array

andreasgan (approved):

PR #298, opened 2023-03-17
Verify emails and edit user profile for non-members

fredrikved:
PR implements gql endpoints for manipulating non-member accounts.


backend/auth0/management.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	token, err := client.GetToken(ctx, client.config.ManagementAudience)
+	if err != nil {
+		return
+	}
+
+	req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", token))
+
+	res, err := http.DefaultClient.Do(req)
```
KillerX, line 37: Default client has infinite timeout, maybe best not to use it.

backend/graph/api/schema/mutations.graphqls
```diff
@@ -26,6 +26,9 @@ type MutationRoot {
     addShowToMyList(showId: ID!): AddToCollectionResult!
 
     removeEntryFromMyList(entryId: UUID!): UserCollection!
+
+    updateUserMetadata(birthData: BirthOptions!, nameData: NameOptions!): Boolean!
+    verifyEmail: Boolean!
```
KillerX, line 31: Is this actually `verify` or is it more `sendVerificationEmail`.

fredrikved: definitely sendVerificationEmail 🙃

backend/auth0/management.go
```diff
@@ -0,0 +1,87 @@
...
+		bodyReader = bytes.NewReader(stringBody)
+	}
+
+	managementUrl := fmt.Sprintf("https://%s/api/v2/", client.config.Domain)
+	req, err := http.NewRequestWithContext(ctx, method, fmt.Sprintf("%s%s", managementUrl, path), bodyReader)
+	if err != nil {
+		return
+	}
+
+	token, err := client.GetToken(ctx, client.config.ManagementAudience)
```
rvanoord, line 35: In general I'm missing "scope" from the code that requests/caches tokens. I assume that means we are getting a token with "all" scopes which isn't really ideal.

backend/graph/api/mutations.resolvers.go
```diff
@@ -451,6 +453,48 @@ func (r *mutationRootResolver) RemoveEntryFromMyList(ctx context.Context, entryI
...
+// UpdateUserMetadata is the resolver for the updateUserMetadata field.
+func (r *mutationRootResolver) UpdateUserMetadata(ctx context.Context, birthData model.BirthOptions, nameData model.NameOptions) (bool, error) {
+	ginCtx, err := utils.GinCtx(ctx)
+	if err != nil {
+		return false, err
+	}
+	u := user.GetFromCtx(ginCtx)
+	if u.IsRegistered() && !u.IsActiveBCC() {
+		displayName := nameData.First + " " + nameData.Last
+		_, err = r.AuthClient.UpdateUser(ctx, ginCtx.GetString(auth0.CtxUserID), auth0.UserInfo{
```
rvanoord, line 465: Can we add some meta information here that identifies the user as a "BCC Media" subscriber?

backend/user/middleware.go
```diff
@@ -215,13 +215,24 @@ func NewUserMiddleware(queries *sqlc.Queries, remoteCache *remotecache.Client, l
 					return i.OrgID
 				})
 			} else {
-				info, err := auth0Client.GetUserInfoForAuthHeader(ctx, ctx.GetHeader("Authorization"))
+				info, err := auth0Client.GetUser(ctx, ctx.GetString(auth0.CtxUserID))
 				if err != nil {
 					return nil, err
 				}
-				u.PersonID = info.Sub
+				u.PersonID = info.UserId
```
rvanoord, line 222: Long term we probably shouldn't use the auth0 ID as the primary key. But that can work for now.

backend/auth0/management.go
```diff
@@ -0,0 +1,87 @@
...
+
+type jobResponse struct {
+	Status    string
+	Type      string
+	CreatedAt string `json:"created_at"`
+	ID        string
+}
+
+// SendVerificationEmail sends a verification email to the specified user
+func (c *Client) SendVerificationEmail(ctx context.Context, sub string) error {
```
rvanoord, line 81: To me it appears that SendVerificationEmail is initiated on the client side via GraphQL (not by the API) and that there isn't a filter in place to ensure it is only sent to BCC Media subscriber users

rvanoord (commented):
Reviewed from a logic point of view (not code quality etc.)

backend/graph/api/mutations.resolvers.go
fredrikved: I thought that was going to be implemented by you guys or?

backend/auth0/management.go
fredrikved: If necessary. We're using all the scopes we have access to.

backend/user/middleware.go
fredrikved: Once we get another key to use, no problem implementing it. We aren't actually using this for much except that.

backend/auth0/management.go
fredrikved: Would that be necessary? Any user that doesn't have a verified email in Auth0 should be able to verify their email as well. Doesn't matter if it's created by the Media client or not.

KillerX (dismissed):

backend/auth0/management.go
rvanoord: ok, if the client is only used within this app

backend/auth0/management.go
rvanoord: How does verification get triggered? is it automatic after first login, or is it something the user invokes?

PR #301, opened 2023-03-24
Implement selecting topics in achievement conditions.

fredrikved:
Topics can be selected individually when creating an achievement condition.

Also implemented ratelimiting as the requests hit the database every time 🙈

### DIFF
--- a/migrations/00161_achieved_items.sql
+++ b/migrations/00161_achieved_items.sql
@@ -1,0 +1,151 @@
+-- +goose Up
+/***********************************************************/
+/*** SCRIPT AUTHOR: Fredrik Vedvik (fredrik@vedvik.tech) ***/
+/***    CREATED ON: 2023-03-24T10:34:47.688Z             ***/
+/***********************************************************/
+
+--- BEGIN CREATE SEQUENCE "public"."achievementconditions_studytopics_id_seq" ---
+
+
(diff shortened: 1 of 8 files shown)


migrations/00161_achieved_items.sql
```diff
@@ -0,0 +1,151 @@
...
+
+--- BEGIN CREATE TABLE "public"."achievementconditions_studytopics" ---
+
+CREATE TABLE IF NOT EXISTS "public"."achievementconditions_studytopics" (
+	"id" int4 NOT NULL DEFAULT nextval('achievementconditions_studytopics_id_seq'::regclass) ,
+	"achievementconditions_id" uuid NULL  ,
+	"studytopics_id" uuid NULL  ,
+	CONSTRAINT "achievementconditions_studytopics_pkey" PRIMARY KEY (id) ,
+	CONSTRAINT "achievementconditions_studytopics_studytopics_id_foreign" FOREIGN KEY (studytopics_id) REFERENCES studytopics(id) ON DELETE SET NULL ,
+	CONSTRAINT "achievementconditions_studytopics_achievem__2e3395b2_foreign" FOREIGN KEY (achievementconditions_id) REFERENCES achievementconditions(id) ON DELETE SET NULL
```
KillerX, line 34: I'm just sort of unsure about all of this being nullable. 
Is this actually needed for some DU stuff or can we cascade it + make it not nullable?

KillerX (approved):
It makes sense to me.

PR #305, opened 2023-03-28
Alternative method for checking for achievements

fredrikved:
We shouldn't need to check for achievements on every pendingachievement request. 

This branch should only check study specific achievements when a task that wasn't completed is completed.


fredrikved:
Started with just implementing a remote cache rate limit solution, but then I figured this is quite weak anyway, as it needs a way to refresh. 

Then I saw that pending achievements shouldn't need to be refreshed on request, only on completion of tasks in this case.

PR #316, opened 2023-04-19
Feat/clear section cache timestamp

fredrikved:
Based on a timestamp pushed to firestore, allow users to retrieve updates live.


backend/cmd/api/loaders.go
```diff
@@ -37,6 +37,9 @@ func getLoadersForRoles(db *sql.DB, queries *sqlc.Queries, collectionLoader *loa
 	rq := queries.RoleQueries(roles)
 
 	ls := &common.FilteredLoaders{
+		GetKey: func() string {
```
KillerX, line 40: I can't quite see the reason as to why this is a function and not just a string.

fredrikved: For interface implementation, but probably not needed, so can just switch to a string.

KillerX (approved):

Issue #325, opened 2023-05-03
My list


mulundapm:
**User story**
As a user I want save episode to a section so that I can easily trace back to these previously saved videos

**Acceptance criteria**
1. Able to have "My list" as a section and able to add this to any page
2. Create a section type which retrieves the items from my list
3. User able to add/ remove items from my list
4. The "my list" section should be refreshed when the user adds/removes items

Design shows my list as a tab but would want to have it as a section in pages before we have more data on how popular it is
[Figma design](https://www.figma.com/file/BMXWzaA93qm8jXKjyPiFOs/BCC-Media-%E2%80%93-Mobile-App-V4?node-id=8822%3A51965&t=m5T6EAvQ6lmdOZPP-1)

fredrikved:
Not sure if it makes sense to support easily adding/removing things from the collection, but it should be doable to create a section type which retrieves the items from my list, in the same way as continue watching.

andreasgan:
Eliezer has fully implemented it as a page now @mulundapm, I completely forgot that you mentioned to me that we should try it as a section first..
What should we do?

mulundapm:
@fredrikved Yes was thinking to have it same way as continue watching. Have edited the task description.

mulundapm:
@andreasgan Does it only require change from Eliezer side? If yes, unfortunately will have to remove it. My apology for not writing it down somewhere.

andreasgan:
done via https://github.com/bcc-code/bcc-media-app/pull/143

PR #331, opened 2023-05-04
Implement my-list as advanced type on collection

fredrikved:
Implement my-list as advanced type on collection.

Allows for pages to be built with mylist as a section instead of a page in itself.

#325

### DIFF
--- a/backend/graph/api/collection-resolver.go
+++ b/backend/graph/api/collection-resolver.go
@@ -8,6 +8,7 @@ import (
 	"github.com/bcc-code/brunstadtv/backend/user"
 	"github.com/bcc-code/brunstadtv/backend/utils"
 	"github.com/bcc-code/mediabank-bridge/log"
+	"github.com/google/uuid"
 	"github.com/samber/lo"
 	"strconv"
 )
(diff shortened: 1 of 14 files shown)


mulundapm:
Hi @fredrikved, are you able to link this to the issue?

KillerX:
Note: I think if you want it to auto link you have to write 

Fixes <ISSUE>

backend/graph/api/collection-resolver.go
```diff
@@ -84,24 +106,42 @@ func sectionCollectionEntryResolver(
...
+			return i.Type == "episode"
+		}), func(i *common.UserCollectionEntry, _ int) uuid.UUID {
+			return i.ItemID
+		})
+		ids, err := ls.EpisodeIDFromUuidLoader.GetMany(ctx, uuids)
+		if err != nil {
+			return nil, err
+		}
+
+		entries = filterWithIds(col, entries, ids)
```
KillerX, line 144: For consideration: Personally I feel that it is not very readable where this begins and ends. I would package the block (or major parts) as a function and call it in the case.

KillerX (dismissed):

backend/graph/api/collection-resolver.go
fredrikved: Agreed.

KillerX (approved):

PR #334, opened 2023-05-05
fix: Do not crash if the "Accept-Language" header is missing

KillerX:


fredrikved (approved):
Not entirely sure what requests fail here, but seems good.

KillerX:
> Not entirely sure what requests fail here, but seems good to avoid malformed header values.

At least cli based downloads of the Sync DB for africa fail :D

PR #351, opened 2023-05-12
feat: Better local db init and management

KillerX:


env.sample
```diff
@@ -0,0 +1,3 @@
+PGUSER=btv
```
fredrikved, line 1: mb shift away from btv as the standard?

guessing this is still included as non-env-variables here and there, so another branch for it in the future?

fredrikved (approved):

env.sample
KillerX: Yeah we should but like you say, it's not fixed everywhere yet.

PR #352, opened 2023-05-15
Feat/downloadables resolution size

KillerX:
This is just the DB/API implementation in order to get that in. <del>The actual implementation will come in a different PR.</del> The implementation has been merged into this branch.


fredrikved (changes requested):
You're missing a property mapping from sqlc.getFilesForAssetsRow to common.File (sqlc/asset-queries.go)

PR #353, opened 2023-05-15
feat(downloadables): Ingest downloadables resolution and size

KillerX:


fredrikved (approved):
Looks ok, but you might want to make audio and subtitlelanguage omitempty in directus.AssetFile, as they will create a language with empty string as code (or use the existing one)... not a part of this pr, but might as well fix it now.

PR #354, opened 2023-05-15
Implement next episodes resolver

fredrikved:
Episode.next will return an array which currently only contains one episode or none.


backend/graph/api/schema/episodes.graphqls
```diff
@@ -46,6 +46,8 @@ type Episode {
     lessons(first: Int, offset: Int): LessonPagination! @goField(forceResolver: true)
     shareRestriction: ShareRestriction! @goField(forceResolver: true)
     inMyList: Boolean! @goField(forceResolver: true)
+
+    next: [Episode!]! @goField(forceResolver: true)
```
andreasgangso, line 50: should be nullable, we may want to stop at the end of a show or playlist, for example.

backend/graph/api/episode-resolver.utils.go
```diff
@@ -0,0 +1,73 @@
...
+		progress, err := r.ProfileLoaders(ctx).ProgressLoader.Get(ctx, utils.AsInt(episodeId))
+		if err != nil {
+			return episodeContext, err
+		}
+		if progress != nil {
+			episodeContext = progress.Context
+		}
+	}
+
+	return episodeContext, nil
```
KillerX, line 28: Personally I like to invert this, i.e.:
```
if { return }
...
```

backend/graph/api/episode-resolver.utils.go
```diff
@@ -0,0 +1,73 @@
...
+}
+
+func (r *episodeResolver) getNextEpisodes(ctx context.Context, episodeId string) ([]int, error) {
+	episodeContext, err := r.getEpisodeContext(ctx, episodeId)
+	if err != nil {
+		return nil, err
+	}
+
+	if episodeContext.CollectionID.Valid {
```
KillerX, line 37: Maybe a quick comment summarizing what's going on here?

KillerX (approved):
Just some stylistic comments, but looks good for me otherwise.

backend/graph/api/schema/episodes.graphqls
fredrikved: It will be empty if there's no next episodes.

andreasgangso: ofc ok

fredrikved:
No, returns only 1 for now.
If we want the user to select between different episodes later, then the array will be larger. For now, there will only ever be 1 or 0 episodes in the array.
@andreasgangso

Issue #344, opened 2023-05-10
Application groups


andreasgangso:
**Problem**
When we have multiple apps, we don't want to share profiles between them (e.g. my-list, continue watching, etc).

We have "applications" and we could split profiles per application.
The problem is that it has fields like "client version", "frontpage", "searchpage", etc. which we may want to distuingish between platforms (web, mobile, and tv). So we would end up with split profiles between platforms.

**Proposed solution**
Introduce Application Groups, example:
- Application Group: bccm
  - Child Applications: bccm-mobile, bccm-tvos, bccm-web
- Application group: kids
  - Child applications: kids-mobile, kids-web
 
We would then split profiles and usergroups per application group, while allowing us to customize frontpage etc per platform.

andreasgan:
@KillerX

KillerX:
Why do we not want to share profiles?

andreasgan:
> Why do we not want to share profiles?

We want to split bcc media users from biblekids users because there are likely different "profiles" using them.
But we need to keep in mind that in the future we may want to share profiles across groups. So maybe make profiles<->appgroups many-to-many

KillerX:
Should only the pages be per application or also the content?

PR #392, opened 2023-05-25
Application groups

fredrikved:
Implement application groups

#344


backend/user/middleware/profiles.go
```diff
@@ -0,0 +1,108 @@
...
+
+func getProfiles(ctx *gin.Context, queries *sqlc.ApplicationQueries, remoteCache *remotecache.Client, user *common.User) ([]common.Profile, error) {
+	key := "profiles:" + user.PersonID
+	if p, ok := profileCache.Get(key); ok && len(p) > 0 {
+		return p, nil
+	}
+
+	lock := utils.Lock(key)
+	lock.Lock()
+	defer lock.Unlock()
```
KillerX, line 55: Do we need to lock on read? Could this be only around "getOrCreate"?

backend/user/middleware/user.go
```diff
@@ -0,0 +1,265 @@
...
+				} else {
+					roles = append(roles, userRoles...)
+				}
+			}
+
+			u.Roles = roles
+
+			ageGroupMin := 0
+			for minAge, group := range user.AgeGroups {
+				// Note: Maps are not iterated in a sorted order, so we have to find the lowed applicable
```
KillerX, line 187:
```suggestion
				// Note: Maps are not iterated in a sorted order, so we have to find the lowest applicable range
```

backend/user/middleware/user.go
```diff
@@ -0,0 +1,265 @@
...
+
+			return u, nil
+		}
+
+		lock, _ := userCacheLocks.LoadOrStore(userID, &sync.Mutex{})
+		lock.Lock()
+		defer func() {
+			lock.Unlock()
+			userCacheLocks.Delete(userID)
+		}()
```
KillerX, line 208: Do you know if there is a functional difference if one were to write:

```
defer lock.Unlock()
defer userCacheLocks.Delete(userID)
```

KillerX (approved):

PR #411, opened 2023-06-02
Directus update and panel extension

fredrikved:
Update Directus to 10.2.

Create a new extension that allows admins to refresh the filter-dataset from a dashboard.


KillerX (approved):
Have you validated permissions to publish to PUBSUB?

fredrikved:
We already use pubsub for search index and some other things, so that should work yes. @KillerX

PR #446, opened 2023-06-27
feat: delayed webviews

### LINKED ISSUE #445: Some prep for new bible study - bmm integration

andreasgangso:
**About the changes**
If the webview code hasnt been injected, but the query param "webview_delayed_type" (flutter/xamarin) is present, make all callHandler() calls wait for the real callHandler code to get injected (signaled by a DOM event: app_webview_ready).

In other words, this makes getAccessToken() wait until the webview is ready before resolving.

Also, the delayed_webview_type is stored in sessionStorage, but can be cleared with query param `?webview_clear=true`. This is just extra safety in case the query param disappears for whatever reason. Shouldn't need to think about it because as far as I know the webview session should never collide with other webview sessions or normal browser sessions.

**Usage example for xamarin**

Append this to the url: `?webview_delayed_type=xamarin`
Inject the following code as soon as possible:
```javascript
window.xamarin_webview = {
    callHandler: (...args) => {
        // implementation
    }
}
window.dispatchEvent(new CustomEvent('app_webview_ready'));
```

Closes #445


fredrikved (approved):
Might want to name the files mainHandler.ts and studyHandler.ts, as snake_case is very dart or?

Otherwise, looks good.

fredrikved (approved):

PR #447, opened 2023-06-27
feat(episode-date): Episode date in title

fredrikved:
Implement Episode dates in title. 

Should be semi-localized, but it doesn't respect the different formats across languages. For example would "2. juni" be written as "2nd of June" or "June 2nd" in English, but this one just translates while persisting the same format.


KillerX (approved):
LGTM. 
Please do a check that this doesn't somehow break the SQLite export.

PR #463, opened 2023-08-03
Replace Directus calls with direct DB calls

fredrikved:
We want to get rid of unnecessary Directus activity, so we rewrite our solutions involving directus to direct db actions.


KillerX (approved):
I think it looks ok. The only concern is that maybe we should wait until monday or even wednesay when the conf is over with the deploy just in case we have a bug? :)

fredrikved:
Wasn't planning to release yet, so wednesday sounds good.

Removed ingest as well, so we'd need to do some testing there so we're sure the flow still works.

backend/crowdin/client.go
```diff
@@ -272,67 +265,223 @@ func (c *Client) Sync(ctx context.Context, d *directus.Handler) error {
...
+					Title:       null.StringFrom(t.Values["title"]),
+					Description: null.StringFrom(t.Values["description"]),
+				})
+			},
+		}, project, directory.ID, crowdinTranslations)
 		if err != nil {
 			return err
 		}
-		err = c.syncGames(ctx, d, project, directory.ID, crowdinTranslations)
+		err = c.syncGames(ctx, TranslationHandler{
```
KillerX, line 475: Would it make sense for this to accept something like:

```
c.syncGames[T any](ctx, func(context, SimpleTranslation, func(ctx, T)), T, ...) 

// Call
c.syncGames(ctx, c.q.UpdateGameTranslation, sqlc.UpdateGameParams{}, ...)
```
Could just look more clean if it is legal in go

KillerX (approved):
A few general comments, nothing blocking from my POV.
I it in general 👍

Issue #456, opened 2023-08-01
Shutdown old.brunstad.tv


KillerX:
- [x] #455 
- [x] Shut down the "azure cloud service" (old import service)
- [x] Remove forward on `/series`
- [x] Remove sync from new
- [x] Remove dns
- [x] Cloud Run disable/scale to 0 and make private so we can be sure no reqs can work
- [x] Dump backup of old DB
- [x] Shut down old DB 
- [x] Decide what to do with AndroidTV

andreasgangso:
> * [ ]  Remove forward on `/series`

not sure what you mean. https://brunstad.tv/series/x should keep working to ensure old links keep working

fredrikved:
> not sure what you mean. https://brunstad.tv/series/x should keep working to ensure old links keep working

At the moment GET requests to this url is sent to the old API due to old player implementations.

KillerX:
Note: DB has been scaled down

KillerX:
Note: Cloud RUN services are now scaling to 0

PR #484, opened 2023-08-28
chore(old.brunstad.tv): Remove loadbalancer redirects and forwards to old

KillerX:
Related: #456


fredrikved:
We are sure that nothing relies on the old web player?

This will remove the old link to btvplayer.js as well.

KillerX:
Yes, I am aware of that. AFAIK It was explorers and kids. @andreasgangso do you know of anything else?

fredrikved (approved):

PR #486, opened 2023-08-29
Data sources

fredrikved:
A separate collection in Directus that can be populated and synced with external sources.

Used only by AssetTimedMetadata for the time being.


migrations/00213_datasources.sql
KillerX, line 1: I think we need the grant for the backup thingy as well.

KillerX (approved):
Just check if we need to grant the Backup user permissions to read tables and seqs

PR #488, opened 2023-08-31
TimedMetadata

fredrikved:
This is the 2nd PR addressing chapters and names/titles for the chapters, making them "type"-safe.

Also implementing "phrases", so we can store and retrieve translations for specific key/values.


backend/graph/api/episodes.resolvers.go
```diff
@@ -156,10 +157,46 @@ func (r *episodeResolver) Chapters(ctx context.Context, obj *model.Episode) ([]*
...
+
 	return lo.Map(metadataItems, func(i *common.TimedMetadata, _ int) *model.Chapter {
+		title := i.Title.Get(languages)
+		phrase, _ := r.Loaders.PhraseLoader.Get(ctx, i.ChapterType.Value)
+		switch i.ChapterType {
+		case common.ChapterTypeSong, common.ChapterTypeSingAlong:
+			if i.SongID.Valid {
+				song, _ := r.Loaders.SongLoader.Get(ctx, i.SongID.UUID)
+				if song == nil {
+					break
```
andreasgangso: Maybe not exclude it if the song is invalid. E.g. "andre sanger". Chapter type can still be song but it just doesn't have a number

backend/graph/api/episodes.resolvers.go
```diff
@@ -156,10 +157,46 @@ func (r *episodeResolver) Chapters(ctx context.Context, obj *model.Episode) ([]*
...
+					}
+					title += song.Title.Get(languages)
+				} else {
+					title = strings.Replace(title, "{{song.title}}", song.Title.Get(languages), -1)
+				}
+			}
+		case common.ChapterTypeSpeech, common.ChapterTypeAppeal, common.ChapterTypeTestimony:
+			if i.PersonID.Valid {
+				person, _ := r.Loaders.PersonLoader.Get(ctx, i.PersonID.UUID)
+				if person == nil {
```
andreasgangso: Same, if person is null, but title is not null, use title instead of excluding?

andreasgangso (changes requested):

backend/graph/api/episodes.resolvers.go
```diff
@@ -156,10 +157,51 @@ func (r *episodeResolver) Chapters(ctx context.Context, obj *model.Episode) ([]*
...
+		case common.ChapterTypeSong, common.ChapterTypeSingAlong:
+			if !i.SongID.Valid {
+				break
+			}
+			song, _ := r.Loaders.SongLoader.Get(ctx, i.SongID.UUID)
+			if song == nil {
+				break
+			}
+			if emptyTitle {
+				title += " - " + song.Title.Get(languages)
```
andreasgangso, line 183: I'd prefer being explicit wherever title is set, instead of appending

andreasgangso (dismissed):
lgtm, but only did a rough logic review

KillerX (dismissed):

PR #568, opened 2023-10-04
Cursors

fredrikved:
Implement cursors so we can enable shuffling.

### DIFF
--- a/backend/graph/api/episode-resolver.utils.go
+++ b/backend/graph/api/episode-resolver.utils.go
@@ -10,6 +10,7 @@ import (
 	"github.com/samber/lo"
 	"gopkg.in/guregu/null.v4"
 	"math/rand"
+	"strconv"
 )
 
 func (r *episodeResolver) getEpisodeContext(ctx context.Context, episodeID string) (common.EpisodeContext, error) {
@@ -35,14 +36,13 @@ func (r *episodeResolver) getEpisodeContext(ctx context.Context, episodeID strin
 	return episodeContext, nil
 }
 
-func (r *episodeResolver) getNextEpisodes(ctx context.Context, episodeID string) ([]int, error) {
+func (r *episodeResolver) getEpisodeQueue(ctx context.Context, episodeID string) ([]int, error) {
 	episodeContext, err := r.getEpisodeContext(ctx, episodeID)
 	if err != nil {
 		return nil, err
 	}
 
 	var episodeIDs []int
-
 	// If the EpisodeContext has a valid CollectionID, use the collectionID to retrieve episodeIDs
 	// else, use the episodes in the season (if any)
 	if episodeContext.CollectionID.Valid {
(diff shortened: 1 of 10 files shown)


fredrikved:
With the current implementation, shuffle/cursor doesn't care about seasons.

Might be the expected behavior at the moment or?

backend/graph/api/episode-resolver.utils.go
```diff
@@ -94,7 +122,7 @@ func (r *episodeResolver) getNextEpisodes(ctx context.Context, episodeID string)
 		}
 		return r.getRelatedEpisodes(ctx, episodeID)
 	}
-	return []int{episodeIDs[index+1]}, nil
+	return []int{next.Current()}, nil
```
andreasgangso, line 125: Can this be next.Keys? On bible kids I need to show a list of the upcoming videos under the video

andreasgangso (changes requested):
1 comment, otherwise lgtm

andreasgangso:
> With the proposed implementation, shuffle/cursor doesn't care about seasons.
> 
> Might be the expected behavior at the moment or?

For bible kids' "play random", I will pick a random available episode in the show client-side, request the episode with `shuffle: true` and play from there. With this implementation it looks like the "next episodes" will only include the current season. So I wont get episodes from other seasons until all the episodes in the season is played. Probably fine for v1.

But I can imagine in the future we will want that "next episodes" list under the bible kids player to be never-ending with 10 episodes or so. For that we would need to have a cursor size (up to a max).

fredrikved:
> > With the proposed implementation, shuffle/cursor doesn't care about seasons.
> > Might be the expected behavior at the moment or?
> 
> For bible kids' "play random", I will pick a random available episode in the show client-side, request the episode with `shuffle: true` and play from there. With this implementation it looks like the "next episodes" will only include the current season. So I wont get episodes from other seasons until all the episodes in the season is played. Probably fine for v1.
> 
> But I can imagine in the future we will want that "next episodes" list under the bible kids player to be never-ending with 10 episodes or so. For that we would need to have a cursor size (up to a max).

Next episodes are from every season in the show.

backend/graph/api/episode-resolver.utils.go
fredrikved: Can expose every next id yes, but that would add additional episodes everywhere. For implementations of auto play the additional episodes are irrelevant.

We might consider making another field for every upcoming episode with pagination.

andreasgangso: yea, but would make more sense the other way I around: move "nextEpisode" into its own field, since this is an array.
Another approach is to add a limit grahql directive if you only care about the next episode.

andreasgangso:
writeup from talk in-person:

- [ ] need consistency for the items in `cursor` and `next` when ctx.cursor == nil
- [ ] add limit directive to `next` with default limit 1

fredrikved:
Implemented consistency across cursor and next, but known limitation atm is if cursor is unspecified, any recursive requests to episodes will be shuffled anew if shuffle is true. So episode.next.cursor != episode.cursor, but episode.next and episode.cursor uses the same cursor. 

Will likely not have any impact in our implementation, but improving it could prove difficult if multiple episodes are requested from different places, so I suggest leaving it as it is.

backend/utils/gin-context.go
```diff
@@ -30,3 +31,41 @@ func GinContextToContextMiddleware() gin.HandlerFunc {
...
+	if ok {
+		r, ok := v.(*T)
+		if !ok {
+			return nil, merry.Errorf("could not cast value to type %T", r)
+		}
+		return r, nil
+	}
+	lock, ok := ginCtx.Get("context-lock-" + key)
+	if !ok {
+		lock = &sync.Mutex{}
```
andreasgangso, line 51: this is not thread safe because two threads can read "context-lock-${key}" at the same time and each create their own lock.

andreasgangso: In general feel using the gin context here is a bit awkward. Can we do it differently? Maybe add some "locks" struct to manage the locks at request creation. (i.e. like the `keyLocks` example from chatgpt below but on the context instead of global). Or just a global map with more unique keys :)


```golang
import (
	"context"
	"sync"
)

var keyLocks = struct {
	sync.Mutex
	m map[string]*sync.Mutex
}{m: make(map[string]*sync.Mutex)}

// GetOrSetContextWithLock gets or sets a value in the context with a lock
func GetOrSetContextWithLock[T any](ctx context.Context, key string, factory func() (*T, error)) (*T, error) {
	ginCtx, err := GinCtx(ctx)
	if err != nil {
		return nil, err
	}

	// Get the specific key lock in a thread-safe manner
	keyLocks.Lock()
	lock, exists := keyLocks.m[key]
	if !exists {
		lock = &sync.Mutex{}
		keyLocks.m[key] = lock
	}
	keyLocks.Unlock()

	lock.Lock()
	defer lock.Unlock()

	v, ok := ginCtx.Get(key)
	if ok {
		r, ok := v.(*T)
		if !ok {
			return nil, merry.Errorf("could not cast value to type %T", r)
		}
		return r, nil
	}

	r, err := factory()
	if err != nil {
		return nil, err
	}
	ginCtx.Set(key, r)
	return r, nil
}
```

> Here's the breakdown of changes:
- Introduced a global keyLocks variable, a struct that contains a map to store a lock for each key and a mutex to guard the access to this map.
- In the GetOrSetContextWithLock function, use this keyLocks to retrieve or create the lock for the given key in a thread-safe manner. This ensures that you have a unique lock for each key and it's created and accessed in a concurrent-safe way.
- Removed the logic that stored the lock in the ginCtx because we're now managing locks with our keyLocks.

andreasgangso (changes requested):

andreasgangso (approved):
lgtm now

backend/utils/cursor.go
```diff
@@ -61,7 +61,17 @@ func (c *Cursor[K]) NextKeys(limit int) []K {
 	if c.CurrentIndex >= len(c.Keys)-1 {
 		return nil
 	}
-	return c.Keys[c.CurrentIndex+1 : lo.Min[int]([]int{c.CurrentIndex + 1 + limit, len(c.Keys)})]
+
+	from := c.CurrentIndex + 1
+
+	to := lo.Min[int](
```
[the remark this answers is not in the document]
fredrikved: in 1.21 yes. haven't updated go.mod... not going to update that here.

PR #570, opened 2023-10-06
Playlists

fredrikved:
Implement playlists.

Playlists are a "wrapper" around collections for usage in sections and collections.

https://github.com/bcc-code/bcc-media-app/issues/259


andreasgangso (changes requested):
I think it makes sense to either:
- Deprecate "Collection" in the graphql api (i.e. `CollectionItem` and in episode context etc)
OR
- Keep the "collection" concept exposed in the API, but have the playlists' items as "collection items". Also, instead of a new type `Item`, use `CollectionItem`.

fredrikved:
> I think it makes sense to either:
> 
>     * Deprecate "Collection" in the graphql api (i.e. `CollectionItem` and in episode context etc)
>       OR
> 
>     * Keep the "collection" concept exposed in the API, but have the playlists' items as "collection items". Also, instead of a new type `Item`, use `CollectionItem`.
> 
> 
> What do you think?

I'm not using a new type. Item is an interface.

We can probably remove some of the collection endpoints and types as those were used by the old API.

andreasgangso:
> I'm not using a new type. Item is an interface. Not all items should be in a playlist either, so only those specified implement the PlaylistItem interface.

Interfaces are also types. Anyway, my point is that CollectionItem and Item are very close in meaning, so one should replace the other. `Item` is a very abstract name best avoided imo.

So I suggest:
```graphql
interface CollectionItem {}
interface PlaylistItem {}
type Show implements CollectionItem {}
type Episode implements CollectionItem & PlaylistItem {}
```

If we really needed it we could do

> We can probably remove some of the collection endpoints and types as those were used by the old API.

Oh I thought `ContextCollection` was using

fredrikved:
Made 2 commits removing old "collection" (used by old api).

fredrikved:
The biggest issue with the approach im doing now is that we cant add additional context based fields to the items in lists without adding properties to those themselves.
 
I don't think that's necessarily bad though, but for example "sort" is gone with this approach.

backend/graph/api/schema/sections.graphqls
```diff
@@ -7,7 +7,7 @@ type ItemSectionMetadata {
     prependLiveElement: Boolean!
 }
 
-interface Section{
+interface Section implements CollectionItem {
```
andreasgangso, line 10: Sections doesnt make sense as collection items

andreasgangso:
> GraphQL has no concept of Collection. Or shouldn't at least. Only the old API uses it.

Episodes use `collectionId` and `ContextCollection` today.

andreasgangso (approved):

backend/graph/api/schema/sections.graphqls
fredrikved: Oh true. It's supported by the backend, though. Just not exposed anywhere (not selectable in admin either). 

Probably a remnant from an older feature.

fredrikved:
> > GraphQL has no concept of Collection. Or shouldn't at least. Only the old API uses it.
> 
> Episodes use `collectionId` and `ContextCollection` today.

True. Kind of magic for the client though. No objects called Collection. 
Could've resolved it from the section... should've done that.

backend/graph/api/schema/sections.graphqls
fredrikved: I can remove it if it's something we don't want to have at all.

andreasgangso: yea probably good to remove it, it would help keep things clean/understandable 👍

andreasgangso (approved):

andreasgangso:
@KillerX can you review this today?

backend/graph/api/schema/pages.graphqls
```diff
@@ -12,15 +12,6 @@ type Page{
     ): SectionPagination! @goField(forceResolver: true)
 }
 
-type Collection {
```
KillerX, line 15: How does this impact the resulting API? I.e. Backwards compat for apps and for Africa?

fredrikved: Did they use Collections?

I've never referenced this in the web code. Only place I know it was used is the old API for the search pages.

andreasgangso: The flutter apps doesnt use it

KillerX: Ok, then it should be fine.

KillerX (approved):

PR #622, opened 2023-10-31
feat(cursor): append related items to next

fredrikved:
Add related items to next when the current queue is ended. 

Basically when the cursor is over, it will add items from the related collection (configured on show level) or episodes with similar tags. 

If there's no related collection and no related episodes, no items will be added when cursor is ended.

Recommended practice is to use the episode cursor retrieved on the original/root episode, and not the cursor inside the next episodes.

### DIFF
--- a/backend/graph/api/episode-resolver.utils.go
+++ b/backend/graph/api/episode-resolver.utils.go
@@ -268,7 +268,7 @@ func (r *episodeResolver) getRelatedEpisodeIDs(ctx context.Context, episodeID st
 	}, []int{})
 	// Uncomment if we want all episodes to be treated equally.
 	// Keep it commented out to weigh episode with more common tags higher.
-	episodeIDs = lo.Uniq(episodeIDs)
+	//episodeIDs = lo.Uniq(episodeIDs)
 	episodeIDs = lo.Filter(episodeIDs, func(i int, _ int) bool {
 		return i != intID
 	})
--- a/backend/graph/api/episode-resolver.utils.go
+++ b/backend/graph/api/episode-resolver.utils.go
@@ -171,6 +171,16 @@ func (r *episodeResolver) getEpisodeCursor(ctx context.Context, episodeID string
 	})
 }
 
+func appendNextEpisodeIDs(keys []int, ids []int) []int {
+	ids = lo.Shuffle(ids)
+	for _, id := range ids {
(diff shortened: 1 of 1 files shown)


KillerX (dismissed):
I personally would not define the lambda there and would rather have made it a full out function but otherwise it seems sane to me.

backend/graph/api/episode-resolver.utils.go
```diff
@@ -180,28 +189,39 @@ func (r *episodeResolver) getNextEpisodes(ctx context.Context, episodeID string,
...
 
-	next := cursor.NextCursor()
-	if next == nil {
-		ids, err := r.getNextFromShowCollection(ctx, episodeID)
+	l := 1
+	if limit != nil {
+		l = *limit
+	}
+
+	keys := cursor.NextKeys(1)
```
andreasgangso, line 197: a bit confusing with the array since it's only 1

backend/graph/api/episode-resolver.utils.go
```diff
@@ -172,6 +171,16 @@ func (r *episodeResolver) getEpisodeCursor(ctx context.Context, episodeID string
 	})
 }
 
+func appendNextEpisodeIDs(keys []int, ids []int) []int {
```
andreasgangso, line 174:
```suggestion
func appendShuffledIDs(keys []int, ids []int) []int {
```

backend/graph/api/episode-resolver.utils.go
```diff
@@ -180,28 +189,39 @@ func (r *episodeResolver) getNextEpisodes(ctx context.Context, episodeID string,
...
+	if len(keys) < l {
+		var ids []int
+		ids, err = r.getNextEpisodeIDsFromShowCollection(ctx, episodeID)
 		if err != nil {
 			return nil, err
 		}
-		if len(ids) > 0 {
-			return ids, nil
+
+		keys = appendNextEpisodeIDs(keys, ids)
```
andreasgangso, line 205: looks like these get shuffled even when the cursor isnt in shuffle mode

andreasgangso (approved):

backend/graph/api/episode-resolver.utils.go
fredrikved: They should be shuffled.

PR #623, opened 2023-11-01
Translations-requirement

fredrikved:
Add boolean on items to determine translation requirement.

When they are false, the items and subitems (unless they have their own toggle) will not be sent to crowdin for translations.


migrations/00230_translations_required.sql
```diff
@@ -0,0 +1,370 @@
+-- +goose Up
+/***********************************************************/
+/*** SCRIPT AUTHOR: Fredrik Vedvik (fredrik@vedvik.tech) ***/
+/***    CREATED ON: 2023-11-01T09:21:19.544Z             ***/
+/***********************************************************/
+
+--- BEGIN ALTER TABLE "public"."lessons" ---
+
+ALTER TABLE IF EXISTS "public"."lessons" ADD COLUMN IF NOT EXISTS "translations_required" bool NULL  ;
```
andreasgangso, line 9: should be `not null`?

migrations/00229_translations_required.sql
```diff
@@ -0,0 +1,468 @@
...
+
+ALTER TABLE IF EXISTS "public"."games" ADD COLUMN IF NOT EXISTS "translations_required" bool NOT NULL DEFAULT true ;
+
+COMMENT ON COLUMN "public"."games"."translations_required"  IS NULL;
+
+--- END ALTER TABLE "public"."games" ---
+
+--- BEGIN ALTER TABLE "public"."shows" ---
+
+ALTER TABLE IF EXISTS "public"."shows" ADD COLUMN IF NOT EXISTS "translations_required" bool NULL DEFAULT true ;
```
andreasgangso, line 57: should be `not null` ?

queries/translations.sql
```diff
@@ -29,8 +30,9 @@ ON CONFLICT (episodes_id, languages_code) DO UPDATE SET title       = EXCLUDED.t
 -- name: ListSeasonTranslations :many
 WITH seasons AS (SELECT s.id
                  FROM seasons s
-                          LEFT JOIN shows sh ON sh.id = s.show_id
-                 WHERE s.status = ANY ('{published,unlisted}')
+                          JOIN shows sh ON sh.id = s.show_id
+                 WHERE s.translations_required
```
andreasgangso, line 34: if I set a field named `translations required´ to false I would expect it to still use existing translations.
Opposite of required is "optional".

andreasgangso (changes requested):

queries/translations.sql
fredrikved: Translations aren't retrieved from this query. This is only for the sync.

fredrikved:
The db migrations is merged already 🙃 

Tweaking some of it now.

PR #640, opened 2023-11-07
chore(db-cleanup): remove unused tables and fix indexes

fredrikved:
Not very easy to see what's going on here, but try to look through the names of what I'm changing and see if you agree or not.

These are the tables (+ related tables) removed in these migrations:
- collection_items
  - deprecated a while back by collection_entries
- categories
- lists

Other than that, the constraints and database types have been fixed to increase the data quality.

I'm expecting this to fail migration when pushing to prod, but I will remove conflicts /fix it manually when needed, unless you have a better suggestion.

### DIFF
--- a/backend/asset/ingest.go
+++ b/backend/asset/ingest.go
@@ -449,8 +449,8 @@ func Ingest(ctx context.Context, services externalServices, config config, event
 
 			for _, l := range audioLanguages {
 				_, err = queries.InsertAssetStreamAudioLanguage(ctx, sqlc.InsertAssetStreamAudioLanguageParams{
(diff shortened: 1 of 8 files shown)


KillerX (commented):
I think it makes sense. 

I think it would make sense to restore a DB locally from a backup and apply the migrations as to make sure they will apply cleanly in PROD. I think it's bad practice if we have migrations that actually can't be applied automatically.

andreasgangso (approved):
lgtm

fredrikved:
Tested the migration locally and added migrations to fix it. Should work fine on prod now.

KillerX (approved):

PR #670, opened 2023-11-20
Shorts

fredrikved:


backend/graph/api/shorts.utils.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+func shortToShort(ctx context.Context, short *common.Short) *model.Short {
+	ginCtx, _ := utils.GinCtx(ctx)
+	languages := user.GetLanguagesFromCtx(ginCtx)
+	return &model.Short{
+		ID:          short.ID.String(),
+		Title:       short.Title.Get(languages),
+		Description: short.Description.GetValueOrNil(languages),
+		Image:       short.Images.GetDefault(languages, common.ImageStyleDefault),
```
andreasgangso, line 85: Just wondering if shorts should use poster or something since it seems `default` means 16:9 thumbnails in practice?

fredrikved: It's supposed to be a resolved field. Moved it there.

backend/graph/api/mutations.resolvers.go
```diff
@@ -165,6 +165,40 @@ func (r *mutationRootResolver) SetEpisodeProgress(ctx context.Context, id string
...
+		return nil, err
+	}
+	if progress != nil {
+		d := 0.0
+		if duration != nil {
+			d = *duration
+		}
+		err = r.GetQueries().SaveVideoProgress(ctx, sqlc.SaveVideoProgressParams{
+			ProfileID: p.ID,
+			Progress:  float32(*progress),
```
KillerX, line 185: Does it make sense to have a 0 progress?
IMO 0 == removal OR "NOP"

fredrikved: nil is removal, 0 could be that you've clicked on the video. 
not a bad thing to track

andreasgangso (approved):

KillerX (approved):

PR #677, opened 2023-11-21
feat(shorts-my-list): add inMyList to shorts and mutation for this

fredrikved:


KillerX (approved):

backend/graph/api/users.utils.go
```diff
@@ -58,3 +66,28 @@ func (r *Resolver) addItemToCollection(ctx context.Context, itemType string, ite
...
+		return false, nil
+	}
+	list, err := r.Loaders.UserCollectionEntryIDsLoader.Get(ctx, utils.AsUuid(myList.ID))
+	if err != nil {
+		return false, err
+	}
+	entryIDs := utils.PointerArrayToArray(list)
+	chunks := lo.Chunk(entryIDs, 20)
+	for _, chunk := range chunks {
+		entries, err := r.Loaders.UserCollectionEntryLoader.GetMany(ctx, chunk)
```
andreasgangso, line 82: Maybe have `UserCollectionEntryIDsLoader` return the ItemIds instead of their "id"?
if only to make the code simpler

andreasgangso (approved):

backend/graph/api/users.utils.go
fredrikved: Maybe in a refactor later.

Not sure why I have an ID for the entries. Could've just been a unique key with collection_id, type and item_id.

andreasgangso: Possibly to enable having the same track multiple times in a playlist. That's a very niche requirement tho, but spotify has it

PR #688, opened 2023-11-27
fix: image size on resize

HarryWangSH:


web/.prettierrc
```diff
@@ -2,5 +2,6 @@
     "trailingComma": "es5",
     "tabWidth": 4,
     "semi": false,
-    "singleQuote": false
+    "singleQuote": false,
+    "endOfLine": "crlf"
```
andreasgangso, line 6:
```suggestion
    "endOfLine": "auto"
```

fredrikved (approved):

fredrikved:
It had to be done on next tick as the component wasn't rendered before the hooks was attached I think, but now you reference the refs in the template, so it should be fine.

PR #747, opened 2024-01-04
feat(roles): compute roles from feature flags

fredrikved:


KillerX (commented):
The general idea seems ok, I'm just wondering if this poses any kind of security risk? I.e. Non-bcc members being able to "just" send a header and get more access?

fredrikved:
It is a risk yes, but if we use it for non-permissive things then it shouldn't matter.

For example only for showing prompts or sections.

backend/utils/feature-flags.go
```diff
@@ -30,21 +31,30 @@ func (f FeatureFlags) Has(key string) bool {
 	return ok
 }
 
+// List returns a list of flags
+func (f FeatureFlags) List() []string {
+	var flags []string
+	for _, flag := range f {
+		r := flag.Key
+		if flag.Variant != "" {
+			r += "-" + flag.Variant
```
andreasgangso, line 40: can we use : to separate variant instead of -? since thats what we are doing in the header

andreasgangso: and i usually call the flags like "something-something"

andreasgangso (approved):
minor thing

backend/utils/feature-flags.go
fredrikved: I just need to stringify it, so it doesn't really matter, but I want to be consistent, so for now its only lowercase and alphanumeric + "-".

fredrikved:
Using `:` as scope now.

`feature-flag:shorts:enabled`

PR #749, opened 2024-01-08
feat(push): allow push to other firebase projects

fredrikved:
Implements push to other apps

Missing:
- infra/permissions for the worker to push notifications to the other project.


fredrikved:
Gave background worker the permissions required now. Just need to test it when this is merged.

KillerX (approved):
Makes sense.

PR #834, opened 2024-02-15
feat: allow livestream to be disabled by filtering on roles

fredrikved:
Add functionality to hide or enable livestream based on specified roles.

In admin, theres a new field to add "livestream roles".


andreasgangso:
nice if we can merge this tomorrow so that i can get it into the next mobile app release

KillerX (approved):
Makes sense. I did not do a deep dive into the migrations but on the surface you are just adding some needed tables, right?

PR #835, opened 2024-02-15
feat: expose asset version

fredrikved:
Expose the updated timestamp for the asset, to identify the asset version in case the user needs to redownload the file.


backend/sqlc/episode-queries.go
```diff
@@ -53,6 +53,7 @@ func (q *Queries) mapToEpisodes(episodes []getEpisodesRow) []common.Episode {
 			SeasonID:              e.SeasonID,
 			AssetID:               e.AssetID,
 			Assets:                assetIDs,
+			AssetVersion:          e.AssetVersion,
```
andreasgangso, line 56: I think it would be slightly more readable if this was `AssetVersion: e.DateUpdated`. I.e. not rename the field in the query

andreasgangso (approved):

backend/sqlc/episode-queries.go
KillerX: I agree. It makes it clear that this is the current mapping.

KillerX:
@oslabbie FYI, this new field may be interesting for you. It basically allows you to detect if the asset on an episode changed. Note that while it is  technically a "date" right now it must be treated as a n opaque string and can change at any time.

andreasgangso (approved):

PR #846, opened 2024-02-23
Add a random factor to shuffle segmented array

fredrikved:
This factor allows for random items to be added from the entire stack to randomize each segment a bit more.

For shorts, this factor will be 1.

This means that if we have 20 new shorts in a segment, 20 other shorts will be added as well (without thinking about segments).


andreasgangso:
Cool!
Sometimes we release larger batches, e.g. 30 shorts. Since that's 2 segments, the user might need to watch 60 shorts to see all the new ones. 
My gut feeling says a factor of 0.5 is better (so you'd need to see 45 shorts to be guaranteed).

I just realized that as the pool grows we might want to surface new-ish shorts more often.
I.e. you open the app, see a cool new short, then a few hours later, if you open the app again, you currently need to watch 200 shorts to see it again...

fredrikved:
> Cool! Sometimes we release larger batches, e.g. 30 shorts. Since that's 2 segments, the user might need to watch 60 shorts to see all the new ones. My gut feeling says a factor of 0.5 is better (so you'd need to see 45 shorts to be guaranteed).
> 
> I just realized that as the pool grows we might want to surface new-ish shorts more often. I.e. you open the app, see a cool new short, then a few hours later, if you open the app again, you currently need to watch 200 shorts to see it again...

Also need to think about the already watched shorts.

A factor of 1 might effectively be closer to 0.5 or even 0 when the user is active.

andreasgangso (approved):
lgtm

KillerX (approved):

PR #914, opened 2024-04-17
feat: initial short logic

andreasgangso:


backend/graph/api/shorts.utils.go
```diff
@@ -223,3 +216,43 @@ func shortToShort(ctx context.Context, short *common.Short) *model.Short {
...
+			break
+		}
+	}
+	if cursor.CurrentIndex == 0 {
+		shortIDs = append([]uuid.UUID{initialID}, shortIDs...)
+	}
+	return shortIDs
+}
+
+// filter out shorts that are watched
```
KillerX, line 233:
```suggestion
// applyWatchedFilter filters out shorts that are watched
```

KillerX (approved):

backend/graph/api/shorts.utils.go
```diff
@@ -187,6 +191,18 @@ func (r *Resolver) getShorts(ctx context.Context, cursor *string, limit *int, in
...
+		for _, s := range shorts {
+			if s.ID == initialUUID {
+				isInResult = true
+				break
+			}
+		}
+		if !isInResult {
+			graphql.AddError(ctx, ErrItemNotFound)
+		}
+	}
```
andreasgangso, line 205: @KillerX i ended up just doing this instead of GetManyStrict. Made more sense to simply add an error if it's not in the result.

backend/loaders/errors.go
```diff
@@ -0,0 +1,11 @@
...
+
+import "github.com/ansel1/merry/v2"
+
+const (
+	ErrorCodeNotFound = "not-found"
+)
+
+func errNotFound(id any) error {
+	return merry.Sentinel(ErrorCodeNotFound, merry.WithValue("id", id))
+}
```
KillerX: This is not a sentinel error, because you are re-creating it. A sentinel error is of the style:

```golang
var errX = merry.Sentinel("blah")

func x() error {
   return merry.Wrap(errX, merry.WithValue("id", "blah"))
}

func main() {
  err := x()
  // Then you can do this
  if errors.Is(err, errX) {
    panic("ERROR")
  }
}
```

KillerX (changes requested):
Looks good, but if we will be doing sentinel errors, let's do them properly.

backend/loaders/errors.go
andreasgangso: ah ok.
This was reverted already so ill just treat the PR as approved 👍

PR #920, opened 2024-04-22
Replace the `mediaitems_view` in some cases.

KillerX:
This allows us to pass in the needed episode_ids and cut the query time by 2 orders of magnitude. 

There are a few consequences:

* SQLc config needed to be upgraded to v2
* In order for SQLc to properly understand the function it needs a active SQL connection
* The function returns are always "nullable" so we need to account for that in the Go code
* Since the returns for the `list` variant is no longer identical we need a separate mapping

The logic is tested but I have not tested it yet with an running API.


KillerX:
Original

image

New

image

backend/sqlc.yaml
```diff
@@ -1,85 +1,116 @@
...
+    gen:
+       go:
+            package: sqlc
+            out: "./sqlc"
+            emit_json_tags: true
+            emit_db_tags: true
+            json_tags_case_style: camel
+            overrides:
+              # For reference, when we need string-arrays (e.g. "usergroups" for a row on the materialized view)
+              # I think I had to do that manually for every column like this:
```
andreasgangso, line 17: lol i think this is a remnant from my POC. can be removed

migrations/00277_optimize_mediaitems_view.sql
```diff
@@ -0,0 +1,106 @@
+-- +goose Up
+
+-- +goose StatementBegin
+CREATE OR REPLACE FUNCTION mediaitems_by_episodes(episodeids int[])
+    RETURNS TABLE
```
andreasgangso, line 5: maybe try OUT params, since this PR is mentioning that: https://github.com/sqlc-dev/sqlc/pull/3175

andreasgangso (approved):
adding the db connection is fine with me although it's ofc nice to avoid it if we can.
the nullability is less comfortable but since it's only that function its not horrible.

- maybe run once locally first.
- try the out param thing, maybe it works without a connection

but i approve, its gefn if you want to merge this

migrations/00277_optimize_mediaitems_view.sql
KillerX: Unfortunately you can't combine them: `[42804] ERROR: RETURN cannot have a parameter in function with OUT parameters`

KillerX: The other alternative would be to create a materialized view for the innermost SELECTS, then we could probably get away with a "normal" view.

KillerX: Now using "setof" option for returning from a function
This takes care both of needing a real DB connection, as well as nullable fields.

backend/sqlc.yaml
```diff
@@ -1,85 +1,114 @@
...
+              # - column: "public.materialized.usergroups"
+              #  go_type:
+              #    import: "github.com/lib/pq"
+              #    type: "StringArray"
```
KillerX, line 19:
```suggestion
```

PR #921, opened 2024-04-22
feat: contributors

andreasgangso:
Basically implemented the resolvers for this:
https://github.com/bcc-code/bcc-media-platform/blob/69565417ddcf4098532dfa823e3c1923a2714c45/backend/graph/api/schema/persons.graphqls


andreasgangso:
@KillerX can you review this? i havent finished testing it and i also need to add some more functionality but ill do it in a different PR

migrations/00278_contributions.sql
```diff
@@ -0,0 +1,622 @@
...
+	MINVALUE 1
+	MAXVALUE 2147483647
+	START WITH 1
+	CACHE 1
+	NO CYCLE;
+
+ALTER SEQUENCE "public"."mediaitems_contributions_id_seq" OWNER TO bccm;
+GRANT SELECT ON SEQUENCE "public"."mediaitems_contributions_id_seq" TO bccm; --WARN: Grant\Revoke privileges to a role can occure in a sql error during execution if role is missing to the target database!
+GRANT USAGE ON SEQUENCE "public"."mediaitems_contributions_id_seq" TO bccm; --WARN: Grant\Revoke privileges to a role can occure in a sql error during execution if role is missing to the target database!
+GRANT UPDATE ON SEQUENCE "public"."mediaitems_contributions_id_seq" TO bccm; --WARN: Grant\Revoke privileges to a role can occure in a sql error during execution if role is missing to the target database!
```
KillerX, line 21: This will fail as there is not `bccm` user. Should be `manager` IIRC. Check some of the other files otherwise

migrations/00278_contributions.sql
```diff
@@ -0,0 +1,622 @@
...
+GRANT USAGE ON SEQUENCE "public"."timedmetadata_styledimages_id_seq" TO bccm; --WARN: Grant\Revoke privileges to a role can occure in a sql error during execution if role is missing to the target database!
+GRANT UPDATE ON SEQUENCE "public"."timedmetadata_styledimages_id_seq" TO bccm; --WARN: Grant\Revoke privileges to a role can occure in a sql error during execution if role is missing to the target database!
+
+COMMENT ON SEQUENCE "public"."timedmetadata_styledimages_id_seq"  IS NULL;
+
+--- END CREATE SEQUENCE "public"."timedmetadata_styledimages_id_seq" ---
+
+--- BEGIN ALTER SEQUENCE "public"."directus_fields_id_seq" ---
+
+ALTER SEQUENCE IF EXISTS "public"."directus_fields_id_seq" START WITH 100000;
```
KillerX, line 49: Random? Sme fix from your side?

migrations/00278_contributions.sql
```diff
@@ -0,0 +1,622 @@
...
+	MINVALUE 1
+	MAXVALUE 2147483647
+	START WITH 1
+	CACHE 1
+	NO CYCLE;
+
+ALTER SEQUENCE "public"."timedmetadata_contributions_id_seq" OWNER TO bccm;
+GRANT SELECT ON SEQUENCE "public"."timedmetadata_contributions_id_seq" TO bccm; --WARN: Grant\Revoke privileges to a role can occure in a sql error during execution if role is missing to the target database!
+GRANT USAGE ON SEQUENCE "public"."timedmetadata_contributions_id_seq" TO bccm; --WARN: Grant\Revoke privileges to a role can occure in a sql error during execution if role is missing to the target database!
+GRANT UPDATE ON SEQUENCE "public"."timedmetadata_contributions_id_seq" TO bccm; --WARN: Grant\Revoke privileges to a role can occure in a sql error during execution if role is missing to the target database!
```
KillerX, line 67: Same as above.

migrations/00278_contributions.sql
```diff
@@ -0,0 +1,622 @@
...
+	MINVALUE 1
+	MAXVALUE 2147483647
+	START WITH 1
+	CACHE 1
+	NO CYCLE;
+
+ALTER SEQUENCE "public"."performances_id_seq" OWNER TO bccm;
+GRANT SELECT ON SEQUENCE "public"."performances_id_seq" TO bccm; --WARN: Grant\Revoke privileges to a role can occure in a sql error during execution if role is missing to the target database!
+GRANT USAGE ON SEQUENCE "public"."performances_id_seq" TO bccm; --WARN: Grant\Revoke privileges to a role can occure in a sql error during execution if role is missing to the target database!
+GRANT UPDATE ON SEQUENCE "public"."performances_id_seq" TO bccm; --WARN: Grant\Revoke privileges to a role can occure in a sql error during execution if role is missing to the target database!
```
KillerX, line 87: Same

migrations/00278_contributions.sql
```diff
@@ -0,0 +1,622 @@
...
+	MAXVALUE 2147483647
+	START WITH 1
+	CACHE 1
+	NO CYCLE;
+
+ALTER SEQUENCE "public"."contributions_id_seq" OWNER TO bccm;
+GRANT SELECT ON SEQUENCE "public"."contributions_id_seq" TO bccm; --WARN: Grant\Revoke privileges to a role can occure in a sql error during execution if role is missing to the target database!
+GRANT USAGE ON SEQUENCE "public"."contributions_id_seq" TO bccm; --WARN: Grant\Revoke privileges to a role can occure in a sql error during execution if role is missing to the target database!
+GRANT UPDATE ON SEQUENCE "public"."contributions_id_seq" TO bccm; --WARN: Grant\Revoke privileges to a role can occure in a sql error during execution if role is missing to the target database!
+
```
KillerX, line 108: Same

migrations/00278_contributions.sql
```diff
@@ -0,0 +1,622 @@
...
+GRANT INSERT ON TABLE "public"."contributions" TO bccm; --WARN: Grant\Revoke privileges to a role can occure in a sql error during execution if role is missing to the target database!
+GRANT UPDATE ON TABLE "public"."contributions" TO bccm; --WARN: Grant\Revoke privileges to a role can occure in a sql error during execution if role is missing to the target database!
+GRANT DELETE ON TABLE "public"."contributions" TO bccm; --WARN: Grant\Revoke privileges to a role can occure in a sql error during execution if role is missing to the target database!
+GRANT TRUNCATE ON TABLE "public"."contributions" TO bccm; --WARN: Grant\Revoke privileges to a role can occure in a sql error during execution if role is missing to the target database!
+GRANT REFERENCES ON TABLE "public"."contributions" TO bccm; --WARN: Grant\Revoke privileges to a role can occure in a sql error during execution if role is missing to the target database!
+GRANT TRIGGER ON TABLE "public"."contributions" TO bccm; --WARN: Grant\Revoke privileges to a role can occure in a sql error during execution if role is missing to the target database!
```
KillerX, line 149: `bccm` does not exist. The api user is called `api`
New things also need to allow `SELECT` by the backup user.

migrations/00279_persons_styledimages.sql
```diff
@@ -0,0 +1,125 @@
...
+GRANT INSERT ON TABLE "public"."persons_styledimages" TO bccm; --WARN: Grant\Revoke privileges to a role can occure in a sql error during execution if role is missing to the target database!
+GRANT UPDATE ON TABLE "public"."persons_styledimages" TO bccm; --WARN: Grant\Revoke privileges to a role can occure in a sql error during execution if role is missing to the target database!
+GRANT DELETE ON TABLE "public"."persons_styledimages" TO bccm; --WARN: Grant\Revoke privileges to a role can occure in a sql error during execution if role is missing to the target database!
+GRANT TRUNCATE ON TABLE "public"."persons_styledimages" TO bccm; --WARN: Grant\Revoke privileges to a role can occure in a sql error during execution if role is missing to the target database!
+GRANT REFERENCES ON TABLE "public"."persons_styledimages" TO bccm; --WARN: Grant\Revoke privileges to a role can occure in a sql error during execution if role is missing to the target database!
+GRANT TRIGGER ON TABLE "public"."persons_styledimages" TO bccm; --WARN: Grant\Revoke privileges to a role can occure in a sql error during execution if role is missing to the target database!
```
KillerX, line 46: `bccm`...

queries/contributions.sql
```diff
@@ -0,0 +1,61 @@
+
+-- name: getContributionCountByType :many
+SELECT type, count(*) as count
+FROM contributions
+WHERE id = ANY (@ids::int[])
+group by type;
```
KillerX, line 6: This one is a bit of a headscratcher for me.

queries/timedmetadata.sql
```diff
@@ -13,8 +36,17 @@ SELECT md.id,
...
+       md.mediaitem_id,
+       COALESCE(images.images, '{}'::json)            AS images
 FROM timedmetadata md
+LEFT JOIN (SELECT simg.timedmetadata_id,
+                      json_agg(json_build_object('style', img.style, 'language', img.language, 'filename_disk',
+                                                 df.filename_disk)) AS images
+               FROM ((timedmetadata_styledimages simg
+                   JOIN styledimages img ON ((img.id = simg.styledimages_id)))
+                   JOIN directus_files df ON ((img.file = df.id)))
+               GROUP BY simg.timedmetadata_id) images ON ((images.timedmetadata_id = md.id))
```
KillerX, line 49: Here the inner query could probably also benefit from having a WHERE.

KillerX (changes requested):

KillerX (changes requested):
I think the it looks good in general, some minor things to check/discuss.

queries/contributions.sql
andreasgangso: lol. its just a count by a set of ids, because im getting ids from getContributionIDsForPersonsWithRoles.
i could include the type in that one and do the counting with go, but this was easier to wrap my head around than the relationloader stuff

KillerX: Ok, but this is not very bulk loader friendly, is it? It will have to run for each contributor separately if I understand it correctly.

andreasgangso: yes, true, good point. I dont think I should spend time to fix it now though, since we are not listing types in lists anyway. It's only used on the contributor page.

backend/graph/api/episodes.resolvers.go
```diff
@@ -18,10 +18,30 @@ import (
...
+	epID, err := r.Loaders.TimedMetadataEpisodeIDLoader.Get(ctx, tmID)
+	if err != nil {
+		return nil, err
+	}
+	if epID != nil {
+		return r.QueryRoot().Episode(ctx, fmt.Sprint(*epID), nil)
+	}
+	return nil, nil
+}
+
```
KillerX, line 44:
```suggestion
// Episode is the resolver for the episode field.
func (r *chapterResolver) Episode(ctx context.Context, obj *model.Chapter) (*model.Episode, error) {
	if obj.Episode != nil && obj.Episode.ID != "" {
		ep, err := r.Loaders.EpisodeLoader.Get(ctx, utils.AsInt(obj.Episode.ID))
		return model.EpisodeFrom(ctx, ep), err
	}
	tmID := utils.AsUuid(obj.ID)
	if tmID == uuid.Nil {
		return nil, nil
	}
	epID, err := r.Loaders.TimedMetadataEpisodeIDLoader.Get(ctx, tmID)
	if err != nil {
		return nil, err
	}
	if epID != nil {
		ep, err := r.Loaders.EpisodeLoader.Get(ctx, *epID)
		return model.EpisodeFrom(ctx, ep), err
	}
	return nil, nil
}
```

I think we should use loaders here otherwise we bypass all aggregation and caching.

backend/graph/api/model/chapter.go
```diff
@@ -0,0 +1,67 @@
...
+	if emptyTitle && phrase != nil {
+		title = phrase.Value.Get(languages)
+	}
+
+	switch tm.ChapterType {
+	case common.ChapterTypeSong, common.ChapterTypeSingAlong:
+		if !tm.SongID.Valid {
+			break
+		}
+		song, _ := loaders.SongLoader.Get(ctx, tm.SongID.UUID)
```
KillerX, line 28: This seems a bit suspicious in the context. In my head the `XFrom` are only data wrangling functions that don't have side effects. Something to look at but don't need to change now.

migrations/00279_persons_styledimages.sql
```diff
@@ -0,0 +1,122 @@
...
+GRANT INSERT ON TABLE "public"."persons_styledimages" TO directus; --WARN: Grant\Revoke privileges to a role can occure in a sql error during execution if role is missing to the target database!
+GRANT UPDATE ON TABLE "public"."persons_styledimages" TO directus; --WARN: Grant\Revoke privileges to a role can occure in a sql error during execution if role is missing to the target database!
+GRANT DELETE ON TABLE "public"."persons_styledimages" TO directus; --WARN: Grant\Revoke privileges to a role can occure in a sql error during execution if role is missing to the target database!
+GRANT TRUNCATE ON TABLE "public"."persons_styledimages" TO directus; --WARN: Grant\Revoke privileges to a role can occure in a sql error during execution if role is missing to the target database!
+GRANT REFERENCES ON TABLE "public"."persons_styledimages" TO directus; --WARN: Grant\Revoke privileges to a role can occure in a sql error during execution if role is missing to the target database!
+GRANT TRIGGER ON TABLE "public"."persons_styledimages" TO directus; --WARN: Grant\Revoke privileges to a role can occure in a sql error during execution if role is missing to the target database!
```
KillerX, line 43: I honestly don't think directus should have this 3.

KillerX (approved):
Looks ok. The notes we can take a look later if we want to.

backend/graph/api/model/chapter.go
andreasgangso: i agree 500%

PR #930, opened 2024-05-06
Feat/statistics tags

andreasgangso:
**Overview**
- removed tagIds ([]int)
- added tags ([]string containing the readable tag.code)

**Question**
What do i do now to update the bigquery tables? Do I go via google console and run a sql query to update the table or do we have migrations or something.

### DIFF
--- a/backend/statistics/bq-schema/episodes.json
+++ b/backend/statistics/bq-schema/episodes.json
@@ -56,6 +56,10 @@
     "type": "JSON"
   },
   {
+    "name": "Tags",
+    "type": "JSON"
+  },
+  {
     "name": "PublicTitle",
     "type": "STRING"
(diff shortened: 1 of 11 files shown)


backend/statistics/bq-schema/episodes.json
```diff
@@ -55,6 +55,10 @@
     "name": "TagIDs",
     "type": "JSON"
   },
+  {
```
andreasgangso, line 58: this file is maybe not used?

KillerX: No, it's not really used. It's not as easy to use as I initally expected and didn't see the value in spending the extra time.

KillerX (approved):

PR #934, opened 2024-05-07
Feat/timedmetadata ingest

andreasgangso:
The main juice is in
asset/timedmetadata.go
asset/timedmetadata.images.go
backend/cmd/jobs/tests/timedmetadata_test.go

I put the test in own package since its closer to an integration test than a unit test. It was much easier and more realistic to run it for real. So its useful for local testing while developing but cant be used in a ci/cd test suite.


KillerX (commented):
I think it makes sense

queries/contributions.sql
```diff
@@ -48,4 +48,8 @@ WHERE
     OR (roles.roles_earlyaccess && @roles::varchar[])
   )
 ORDER BY
-  m.published_at DESC;
\ No newline at end of file
+  m.published_at DESC;
+
+-- name: InsertContribution :exec
+INSERT INTO "public"."contributions" (person_id, "type", mediaitem_id, timedmetadata_id)
+VALUES (@person_id::uuid, @type, sqlc.narg('mediaitem_id')::uuid, sqlc.narg('timedmetadata_id')::uuid);
```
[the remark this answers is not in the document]
andreasgangso: not same thing. ::uuid is not nullable. `narg` makes it nullable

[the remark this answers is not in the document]
andreasgangso: https://docs.sqlc.dev/en/stable/howto/named_parameters.html#nullable-parameters
"there is no nullable equivalent of the @-syntax"

backend/asset/timedmetadata.images.go
```diff
@@ -0,0 +1,76 @@
...
+	"github.com/ansel1/merry/v2"
+	"github.com/bcc-code/bcc-media-platform/backend/files"
+	"github.com/bcc-code/mediabank-bridge/log"
+	"golang.org/x/sync/errgroup"
+)
+
+// ingestImagesFromS3 downloads images from S3 and uploads them to the platform
+func ingestImagesFromS3(ctx context.Context, imagePaths []string, services externalServices, config config) (map[string]string, error) {
+	s3client := services.GetS3Client()
+	tempDir, err := os.MkdirTemp(config.GetTempDir(), "timedmetadata")
```
KillerX, line 18: Is there a specific reason we are not using `os.TempDir()` but are instead introducing a whole new config var?

backend/asset/timedmetadata.images.go
```diff
@@ -0,0 +1,76 @@
...
+			}
+
+			imageId, err := uploadFileToPlatform(ctx, services, localPath)
+			if err != nil {
+				return merry.Wrap(err)
+			}
+			mu.Lock()
+			imageIDs[i] = *imageId
+			mu.Unlock()
+			return nil
```
KillerX, line 48: Is there a direct need to download the file to disk? It seems like the stream could be passed directly from the "dowload" to the "upload" functions.

KillerX (approved):

backend/asset/timedmetadata.images.go
KillerX, line 1: Please don't introduce a 3rd file naming :D `timedmetadata_images.go` is the most go-like

KillerX (approved):
Some comments for potential improvements, but lgtm in general. Up to you if you want to do something about it.

backend/asset/timedmetadata.images.go
```diff
@@ -0,0 +1,76 @@
...
+			})
+			if err != nil {
+				return merry.Wrap(err)
+			}
+
+			imageId, err := uploadFileToPlatform(ctx, services, localPath)
+			if err != nil {
+				return merry.Wrap(err)
+			}
+			mu.Lock()
```
KillerX, line 45: Why do we need a lock here? I it not safe if we know that we are writing into different keys?

backend/asset/timedmetadata.images.go
andreasgangso: nope. changed to do that

PR #942, opened 2024-05-30
Change chapter_type to content_type

KillerX:
The GQL api was not changed but now uses the new content_type. 

There is still a unclear situation when a timed metdata is related to many different episodes. This appears to only be the case at the moment with "Placeholder" video, and makes little logical sense anywhere else, so I have not done anything specific for this case, as the placeholder should not have chapters anyway.


KillerX:
@andreasgangso added the GQL endpoints for content type. The other `type` is still present in order to not break things, even if we decide not to use it in the future.

#942

andreasgangso (approved):

PR #943, opened 2024-05-31
Exploration into End-To-End tests

KillerX:
# First up, this is *not* an PR I expect to be merged!!

This is an exploration into what the is the minimum amount of work needed to get some kind of end-to-end going where the data is in a controlled state and as much as possible is "built in", I.e. you don't need manual intervention between tests.

Observations/questions so far:

* GQL is not the most friendly language in this context
  * Most of the "GQL client libraries" handle the actual connection and request internally so it is not (possible?) easy to get to the `req` struct that needs to be passed to the test recorder. Alternative is to start a server listening on a port in a go routine and stop it via context deadline.
* It would take some effort to create a *maintainable* set of data that can be queries
* The performance is not horrible considering all migrations etc, ~6 seconds
* Considering most of the DB is RO from the GQL perspective, we should just clean parts that are R/W before a test and leave the rest in tact. Then have a "button" to re-generate the test DB

Please let me know if you have any smart ideas to try/test.

I will also do a quick exploration with mocking the `queries` struct to see if that is an easier path. It's not end to end but it should still be a helpful harness for developing and reproducing bugs (as long as they are not in a query directly).


kkuepper:
When looking at the [test pyramid](https://martinfowler.com/articles/practical-test-pyramid.html#TheTestPyramid) those tests are probably somewhere between UI tests and service tests, right?

andreasgangso:
I think this is a good start. no smart ideas to add

KillerX:
> When looking at the [test pyramid](https://martinfowler.com/articles/practical-test-pyramid.html#TheTestPyramid) those tests are probably somewhere between UI tests and service tests, right?

Yes, I think that makes sense.

PR #954, opened 2024-06-06
feat: imagor video

andreasgangso:
Havent set up the service yet but it works locally


backend/asset/s3_videos.go
```diff
@@ -0,0 +1,70 @@
...
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/ansel1/merry/v2"
+	"github.com/aws/aws-sdk-go-v2/service/s3"
+	"github.com/aws/aws-sdk-go-v2/service/s3/types"
+)
+
+func getSignedAudiolessVideoURLForAssetID(context context.Context, services externalServices, config config, assetID int32) (*string, error) {
```
KillerX, line 14: Why `Audioless`? I.e. Do we care?

backend/asset/s3_videos.go
```diff
@@ -0,0 +1,70 @@
...
+		Bucket: config.GetStorageBucket(),
+		Prefix: &storagePath.String,
+	})
+	if err != nil {
+		return nil, merry.Wrap(err)
+	}
+
+	if len(objects.Contents) == 0 {
+		return nil, merry.New("no objects found", merry.WithUserMessage("no objects found for assetID: "+string(assetID)))
+	}
```
KillerX, line 37: I think you can drop this section, because the loop doesn't care if the len == 0 and you check afterwards anyways.
```suggestion
```

backend/asset/timedmetadata_images.go
```diff
@@ -51,3 +54,39 @@ func ingestImagesFromS3(ctx context.Context, imagePaths []string, services exter
...
+	if err != nil {
+		return nil, merry.Wrap(err)
+	}
+	defer image.Close()
+
+	fs := services.GetFileService()
+	uuid, err := uuid.NewV7()
+	if err != nil {
+		return nil, merry.Wrap(err)
+	}
```
KillerX, line 79:
```suggestion
	uuid := uuid.New()
```

backend/imagor/generate.go
```diff
@@ -0,0 +1,88 @@
...
+	} else {
+		path = imagorpath.Generate(params, imagorpath.NewDefaultSigner(i.signingKey))
+	}
+	imagorUrl := fmt.Sprintf("%s/%s", i.baseUrl, path)
+	req, err := http.NewRequest("GET", imagorUrl, nil)
+	if err != nil {
+		return nil, merry.Wrap(err)
+	}
+
+	client := http.Client{}
```
KillerX, line 77: Note: Default go http has no sane timeout set (https://medium.com/@nate510/don-t-use-go-s-default-http-client-4804cb19f779). This is one of the reasons I tend to reach for https://github.com/go-resty/resty . It also handles retries and stuff like that. Don't have to use it but at the very least a timeout should be added.

docker-compose.yml
```diff
@@ -19,6 +19,11 @@ services:
     image: redis
     ports:
     - "6378:6379"
+  imagor:
+    image: shumc/imagorvideo
+    ports:
+    - "8005:8000"
+    command: ["-imagor-unsafe"]
```
KillerX, line 26: Does unsafe relate to no https or something else?

backend/asset/s3_videos.go
andreasgangso: Well just wanted it to be clear so you dont expect audio

docker-compose.yml
andreasgangso: refers to not needing a signing key

PR #956, opened 2024-06-07
feat: switch out imagor with a new service (videomanipulator)

andreasgangso:


backend/Dockerfile.ffmpeg
KillerX, line 1: Is this needed? It seems like it's integrated in the other Dockerfile.

KillerX (approved):

backend/Dockerfile.ffmpeg
andreasgangso: ah thanks. yea this is the one being used, the other one was junk

PR #982, opened 2024-07-11
feat: skip to chapter

andreasgangso:


backend/graph/api/episodes.resolvers.go
```diff
@@ -199,31 +201,38 @@ func (r *episodeResolver) Files(ctx context.Context, obj *model.Episode, audioLa
 
 // Chapters is the resolver for the chapters field.
 func (r *episodeResolver) Chapters(ctx context.Context, obj *model.Episode) ([]*model.Chapter, error) {
-	i, err := r.Loaders.EpisodeLoader.Get(ctx, utils.AsInt(obj.ID))
-	if err != nil || !i.AssetID.Valid {
+	return r.getChapters(ctx, obj.ID)
+}
+
+// SkipToChapter is the resolver for the skipToChapter field.
```
kkuepper, line 207: can't we use better method names, so that this comment is not necessary?

backend/graph/api/episodes.resolvers.go
```diff
@@ -199,31 +201,38 @@ func (r *episodeResolver) Files(ctx context.Context, obj *model.Episode, audioLa
...
-			return nil, err
+	slices.SortStableFunc(chapters, func(i, j *model.Chapter) int {
+		return cmp.Compare(i.Start, j.Start)
+	})
+	var chapter *model.Chapter
+	for _, c := range chapters {
+		if c.Start > obj.Duration/2 {
+			break
+		}
+		if c.ContentType != nil && (c.ContentType.Code == common.ContentTypeTheme.Value || c.ContentType.Code == common.ContentTypeInterview.Value || c.ContentType.Code == common.ContentTypeSpeech.Value) {
```
kkuepper, line 225: not sure what theme is. And not sure whether I should use time to understand what it means. Probably fine.

kkuepper: the rest of the logic looks plausible.

backend/graph/api/episodes.resolvers.go
kkuepper: I don't know enough about go. But `skipToChapter: Chapter @goField(forceResolver: true)` from episodes.graphqls makes me wonder if this comment is inaccurate.

andreasgangso: its automatic via codegen x)

kkuepper (approved):

PR #985, opened 2024-07-15
feat: hide web section for ipads on newer versions

andreasgangso:


KillerX (changes requested):
Why do we want that? Why only iOS?
And if we do want that, why does the backend decide on some "arbitrary" data? Should this not be something the frontend signals, or some cleaner way of identifying who should get it?

The code right now appears very "hacky" and there will be imo very hard to understand what's going on down the line.

andreasgangso:
Yea this is old just delete it.
Was to hotfix a bug, but we didnt really need to

Sent from Outlook for iOS<https://aka.ms/o0ukef>
________________________________

Issue #969, opened 2024-06-17
Shorts algorithm


mulundapm:

KillerX:
Note: I updated the formulas so that shorts that have `0` interactions still get a score and not a NULL.

KillerX:
```
// getShort gets a list of paginated shorts
// 
// The current logic is:
// * Get the shorts grouped by publishing week sorted DESC by `YEAR-WEEK`
// * Sort them by score (descending) within each group (week) <--- NEW
// * The shorts get shuffled (See ShuffleSegmentedArray)
// * Any watched shorts are removed
// * If there is a forced "Initial Short" it gets prepended
// * If we do not have enough shorts we clear the progress and repeat the process, appending on the end
// * Return a cursor with the data
```

@mulundapm @kkuepper See above for a rough description about the current algorithm

ShuffleSegmentedArray: https://github.com/bcc-code/bcc-media-platform/blob/d2fcafb3f75d1855ee0bc155a467e92a6e3adca0/backend/utils/shuffle.go#L22
Tests that maybe visualize it better: https://github.com/bcc-code/bcc-media-platform/blob/d2fcafb3f75d1855ee0bc155a467e92a6e3adca0/backend/utils/shuffle_test.go#L11

KillerX:
New shorts get 4 points, degrades by 0.5 pts per day. Stops at 0 "boost".

KillerX:
The unleash flag required to enable the new algorithm is `shorts-with-scores`.

KillerX:
Im 99% sure the cause is that you had the old query cached. The new query results in a different data type, so we can't use the same cache key. Fix: #1011

KillerX:
Fix deployed.

mulundapm:
Average score of well performing shorts are at 4, if an addition of 4 is added as recency boost, it would only show recently published shorts basically for the first 4-5 days. 

Noticed there is a disadvantage of having this, we create shorts in batches from a same assets. Within 10 new shorts, 3-4 shorts are from same film, which could seem very repetitive if they show up as first 10 shorts. 

Could we add a randomize in recency boost? Would this help? @KillerX 

e.g. for each short, score = engagement score + recency boost * random of 0.1-1

PR #1008, opened 2024-09-20
Add Scores to shorts alogrithm

KillerX:
This is the smallest change possible to start considering scores. We can discuss if we want to do more than that.

### DIFF
--- a/backend/graph/api/shorts.utils.go
+++ b/backend/graph/api/shorts.utils.go
@@ -175,16 +175,6 @@ func (r *Resolver) clearShortsProgress(ctx context.Context, p *common.Profile) e
 	return nil
 }
 
-// getShort gets a list of paginated shorts
-//
-// The current logic is:
-// * Get the shorts grouped by publishing week (based on what user has access to)
-// * Internally sort them by score (descending)
-// * The shorts get shuffled (See ShuffleSegmentedArray)
-// * Any watched shorts are removed
-// * If there is a forced "Initial Short" it gets prepended
-// * If we do not have enough shorts we clear the progress and repeat the process, appending on the end
-// * Return a cursor with the data
 func (r *Resolver) getShorts(ctx context.Context, cursor *string, limit *int, initialID *string) (*model.ShortsPagination, error) {
 	p, err := getProfile(ctx)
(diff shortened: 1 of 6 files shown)


KillerX:
@kkuepper updated query (and the needed machinery)

The algo now:

* Calculate a boost (5 pts - 0.5pts/day, min 0)
* boost + score == final_score
* Order by final score DESC
* Remove watched
* Shuffle internally in blocks of 5

backend/graph/api/shorts.utils.go
```diff
@@ -161,6 +175,16 @@ func (r *Resolver) clearShortsProgress(ctx context.Context, p *common.Profile) e
 	return nil
 }
 
+// getShort gets a list of paginated shorts
+//
+// The current logic is:
+// * Get the shorts grouped by publishing week (based on what user has access to)
```
kkuepper, line 181: is the comment outdated?

backend/unleash/flags.go
```diff
@@ -0,0 +1,7 @@
+package unleash
+
+// ShortsWithScores enables shorts sorting by scores
+//
+// Started: 20.9.2024
+// To be Removed Latest: 20.12.2024
+const ShortsWithScores = "shorts-with-scores"
```
kkuepper, line 7: has this been added to unleash?
If not, you should write the id in #969

queries/shorts.sql
```diff
@@ -12,6 +12,25 @@ WHERE s.status = 'published'
 GROUP BY week
 ORDER BY week DESC;
 
+-- name: ListSegmentedShortIDsForRolesWithScores :many
+SELECT s.id,
+       -- We need a date and if we do not have a published_at date, we need to assume that the created date is when it was published
+       EXTRACT(DAY FROM current_date - COALESCE(mi.published_at, mi.date_created)) age,
+
+       -- For 10 days the shorts is boosted. It starts with a 5 points boost, and the boost "degrades" by 0.5
```
kkuepper, line 20: boosting it by 5 points, means that the first 2 days, most shorts shown will be the new ones. That has the nice side-effect that we get feedback about the short a lot quicker than before because ~150 people will have seen the short after 1 day.

kkuepper (approved):

PR #1014, opened 2024-09-27
Return kids-mobile as application code for old app versions

mulundapm:


backend/applications/middleware.go
```diff
@@ -33,6 +34,14 @@ func ApplicationMiddleware(
 	}
 }
 
+// Support app version with old application name kids
```
kkuepper, line 37: In general it's better to only comment why you're doing something and not the what.
For the What it's better to refactor the code and make it easier to understand.

kkuepper (approved):

andreasgangso (changes requested):

kkuepper (approved):

PR #1016, opened 2024-09-28
chore: update directus

andreasgangso:
The migrations
- Directus added a new table "access policies". Permissions now point to policies instead of roles. Some of the fields that were previously on "role" got moved to policies. The policies got the same uuid as the roles, so its a simple operation to move the relations (`UPDATE "public"."directus_permissions" SET "role" = "policy";`).
- `directus_access` is a new junction table to connect roles (or users) to policies.

Other things
- Move modules into the extensions folder directly. This was a breaking change in v10 [changelog](https://github.com/directus/directus/blob/e8d96d47066613c5a026f7af52cbe9e961ac0017/docs/releases/breaking-changes.md?plain=1#L130)
- Updated all the cms and extensions' npm packages
- Just some makefile and documentation tweaks that were convenient to do while I was working

Risks
- Updated major versions for so although things build there could be changes in behavior.
- Hard to confirm locally if all the extensions work as they should. The most critical extension to keep an eye out for is the pubsub hooks.
- We might start seeing some errors in admin because directus has become stricter: "[Requests for Missing Fields Now Fail](https://github.com/directus/directus/blob/e8d96d47066613c5a026f7af52cbe9e961ac0017/docs/releases/breaking-changes.md?plain=1#L50)". I noticed and fixed one such case where a directus_fields row for `timedmetadata` still had `timestamp` (we renamed it to `seconds` at some point).


andreasgangso:
noticed some issues with the ui extensions

KillerX (approved):

PR #1021, opened 2024-10-08
Add Elasticsearch support

KillerX:
It's enabled using a feature flag.
Indexing runs for both Algolia and ES.
Elastic tests can be run using `make test.elastic` from the `backend` folder.

Every search result has a `searchProvided` field available (if requested in the query) that identifies the source of the results as "algolia" or "elasticsearch", for easier testing/verification where we are getting the results from.


KillerX:
I think this can be reviewed. The query is working as expected in terms of filtering, but it's probably missing a lot of fine tuning.

KillerX:
@kkuepper Is there a conflict if I put `bccm-*` indexes into the same elastic cluster? Does BMM have a limit by index or does it search all indexes?

kkuepper:
> @kkuepper Is there a conflict if I put `bccm-*` indexes into the same elastic cluster? Does BMM have a limit by index or does it search all indexes?

I think that's fine. BMM only checks the specific index it also creates.

backend/search/query_test.go
```diff
@@ -0,0 +1,60 @@
...
+		},
+	)
+
+	assert.NoError(t, err)
+	assert.NotNil(t, res)
+	assert.Equal(t, 6, res.HitCount)
+
+	// If you want to see the results in detail in terminal
+	//spew.Dump(res)
+}
```
KillerX, line 60: @kkuepper can you confirm that this makes sense to you.

kkuepper: I guess. It's the beginning of a test.

PR #1026, opened 2024-10-10
Implement "declumping" of shorts

KillerX:


backend/graph/api/shorts.utils_test.go
```diff
@@ -0,0 +1,128 @@
...
+					ID:              uuid.MustParse("00000000-0000-0000-0000-000000000004"),
+					ParentEpisodeID: null.IntFrom(1),
+				},
+			},
+			expected: []uuid.UUID{
+				uuid.MustParse("00000000-0000-0000-0000-000000000001"),
+				uuid.MustParse("00000000-0000-0000-0000-000000000004"),
+			},
+			declumpingTreshold: 3,
+		},
```
KillerX, line 68: What would you guys expect as the result here? Is it what I made or would you expect only `00000000-0000-0000-0000-000000000001` to be in the result?

backend/common/models.go
```diff
@@ -148,3 +148,15 @@ type Redirect struct {
...
 }
+
+// ShortIDWithMeta
+//
+// Age - Age of the short in days
+// ParentEpisodeID - ID of the full episode where the short was extracted from
+// FinalScore - Final score of the short, with all adjustments
+type ShortIDWithMeta struct {
+	ID              uuid.UUID
+	Age             string
```
sifferhans, line 159: Isn't it better to use a timestamp here instead?

backend/graph/api/shorts.utils_test.go
sifferhans: I think it makes sense if we think about it like the threshold is how many until the next one, instead of how many between this and the next one.

sifferhans (approved):
Looks mostly good, only a simple question regarding the `ShortIDWithMeta` struct

backend/common/models.go
KillerX: In this case INT would be correct. I'll fix both the type and the name `AgeInSeconds` or something along those lines

PR #1049, opened 2024-10-30
feat(feature-flags): Add header to the response that indicates if a f…

KillerX:
…eature flag was used

Example response herder: 

X-Feature-Flags: shorts-with-scores2:enabled

### DIFF
--- a/backend/utils/feature-flags.go
+++ b/backend/utils/feature-flags.go
@@ -76,3 +76,8 @@ func GetFeatureFlags(ctx *gin.Context) FeatureFlags {
 
 	return featureFlags
 }
+
+// ReportFlagActivation reports flag activation as a header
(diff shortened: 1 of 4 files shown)


backend/utils/feature-flags.go
```diff
@@ -76,3 +76,8 @@ func GetFeatureFlags(ctx *gin.Context) FeatureFlags {
 
 	return featureFlags
 }
+
+// ReportFlagActivation reports flag activation as a header
+func ReportFlagActivation(ctx *gin.Context, flag string, variant string) {
```
sifferhans, line 81: Variants can theoretically be undefined in Unleash, right? Based on what type of feature flag

sifferhans (approved):

backend/utils/feature-flags.go
KillerX: Tanks, handling that now

PR #1058, opened 2024-11-06
feat(analytics): add video played event

sifferhans:


web/src/components/EpisodeViewer.vue
```diff
@@ -171,6 +174,15 @@ const load = async () => {
         player.value.on('ended', () => {
             emit('next')
         })
+        player.value.on('play', () => {
+            analytics.track('video_played', {
+                videoId: props.episode.id,
+                referenceId: referenceId,
+                data: {
+                    // Whatever data we want to send
```
KillerX, line 182: For QC maybe it would be smart to add `ageGroup` right away?
Then we could join the tables and compare the 2 felds to make sure we are matching things as expected.
I'm in addition to that RS Session ID should also match.

web/src/services/analytics/events.ts
```diff
@@ -109,4 +109,9 @@ export type Events = {
         elementId?: string
         meta?: apiObject
     }
+    video_played: {
+        videoId: string
+        referenceId?: string
+        data?: Record<string, unknown>
+    }
```
KillerX, line 116: Can you also update https://github.com/bcc-code/bcc-media-adr/blob/master/02-analytics/technical.md

KillerX (approved):

web/src/services/analytics/events.ts
sifferhans: https://github.com/bcc-code/bcc-media-adr/pull/30

web/src/components/EpisodeViewer.vue
sifferhans: Added it to all `analytics.track` calls 👍

PR #1074, opened 2024-12-04
Upgrade video.js in `bccm-video-player`

sifferhans:
- subtitles on live now works
- fixed a bug where subtitles were not set correctly when player was initialized


KillerX:
@oslabbie does this cause any issues for you guys?

KillerX (commented):
LGTM, pending SA confirmation that it does not cause issues for them

oslabbie:
@KillerX, don't think this will cause an issue for us, as long as there is not changes on the sqlite db.

PR #1084, opened 2024-12-17
create index as part of the test

kkuepper:


backend/search/query_test.go
```diff
@@ -22,18 +27,33 @@ func TestMain(m *testing.M) {
...
 
+	mapping := types.NewNestedProperty()
+	mapping.Properties = map[string]types.Property{
+		"total": types.NewIntegerNumberProperty(),
+		"free":  types.NewIntegerNumberProperty(),
+		"used":  types.NewIntegerNumberProperty(),
+	}
+
+	client.Indices.Create(testIndexName).Do(ctx)
+	client.Indices.PutMapping(testIndexName).Request(&putmapping.Request{Properties: map[string]types.Property{
```
kkuepper, line 50: what does the index look like? Can you point me to the definition of the index?

kkuepper: @KillerX

kkuepper: The idea is to get something similar to what we have in BMM: https://github.com/bcc-code/bmm-api/blob/develop/BMM.Api.Infrastructure.Tests/ElasticSearch/SearchTrackTests.cs

- create an index for the test
- fill it with the test data that is needed for this test
- run actual test
- delete everything

KillerX: https://github.com/bcc-code/bcc-media-platform/blob/master/backend/search/fields.go#L36

backend/search/query.go
```diff
@@ -243,6 +243,9 @@ func doElasticSearch(ctx context.Context, client *elasticsearch.TypedClient, que
 	if query.Type != nil {
 		indexName = typeToIndexMap[*query.Type]
 	}
+	if indexNameParameter != "" {
+		indexName = indexNameParameter
+	}
```
KillerX, line 248: The index is set based on what you are queriying. You should not be able to query for episodes and spec to search the season index.
If that is needed for testing then initialization of the client should accept index config.

Issue #1083, opened 2024-12-17
Content of cursor is same as next cursor for shorts


mulundapm:
Shorts are repeating after roughly 10 scrolls on algorithm

KillerX:
Fact confirmed so far:

* The repeated shorts are coming from the API
* The cursors are used correctly (i.e. the cursor does not get reset somewhere)
* Currently unable to repro the issue locally

mulundapm:
This is still an occurring issue.

mulundapm:
This is still reproducible for me

In addition, I watched shorts, quit app, open the app again and go to short, i'm getting about the same shorts as previous session. 

tested with both score enabled and disabled.

sifferhans:
Nevermind, they start to duplicate again after ~20 shorts are scrolled.

KillerX:
@sifferhans ready for some pair programming tomorrow. There is probably some nuance I am missing somewhere.

sifferhans:
@KillerX Yes, probably nice to look at it with four eyes

PR #1085, opened 2024-12-20
Rewrite of the shuffling system

KillerX:
I have rewritten the shuffling system completely as it was very complex and not useful at all when looking at how we have begun to approach the shorts algo now.

I have also added tests to validate that the pagination logic is correct.

I do not know 100% that this will solve #1083 but it basically came about as I tried to understand what is going on.

The core of the changes is in backend/utils/shuffle.go (significant simplification).
Most of the other stuff is adjustments to the new function signature, tools that I built for debugging, and tests.


sifferhans (approved):
lgtm, cannot see any major red flags

PR #1108, opened 2025-02-07
Implement filtering by topic on lessons completed conditions

KillerX:
`CheckCompletedLessonAchivements` is the core of the implementation. 

The rest is mostly support code for testing and SQL queries.


KillerX:
Note: There is a migration missing that exposes an existing field in the UI. Coming on Monday.

sifferhans (approved):

PR #1175, opened 2025-04-25
feat: add bmm streak tracking to web

sifferhans:
Adds a call to BMM's `/Statistics/watched` endpoint when watching a video.

There are two scenarios where we call the endpoint:
- The video current time is at > 50% of the video duration, and has been for 10 progress "ticks"  
- The video is finished

### DIFF
--- a/web/src/components/EpisodeViewer.vue
+++ b/web/src/components/EpisodeViewer.vue
@@ -256,6 +256,9 @@ const checkProgress = async (force?: boolean) => {
     }
 }
 
+// This is a very naive v1.0 of the BMM streak tracking.
+// With this implementation, we have basically no way of preventing the
+// user from skipping forward to the end, and therefore getting the streak point.
 let hasSentHalfwayWatchedEvent = false
 function onVideoProgress(
     currentTime: number,
--- a/web/src/components/EpisodeViewer.vue
+++ b/web/src/components/EpisodeViewer.vue
@@ -20,6 +20,9 @@ import { languages } from '@/services/language'
 import { currentApp } from '@/services/app'
 import { languageTo3letter } from '@/utils/languages'
 import { generateUUID } from '@/utils/uuid'
+import { BMM } from '@/services/bmm'
+import { getOperatingSystem } from '@/utils/userAgent'
+import { ProcessWatchedCommandEvent } from '@bcc-code/bmm-sdk-fetch'
 
 const { isAuthenticated } = useAuth0()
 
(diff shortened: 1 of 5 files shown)


web/src/utils/userAgent.ts
```diff
@@ -0,0 +1,15 @@
+export function getOperatingSystem() {
```
kkuepper, line 1: currently, we're using lowercase versions of OS. So ios or android.

web/src/components/EpisodeViewer.vue
```diff
@@ -250,6 +256,34 @@ const checkProgress = async (force?: boolean) => {
     }
 }
 
+let playedOverHalfCounter = 0
```
kkuepper, line 259: I don't understand why and how this is used

web/src/components/EpisodeViewer.vue
```diff
@@ -250,6 +256,34 @@ const checkProgress = async (force?: boolean) => {
...
+function onVideoProgress(
```
kkuepper, line 260: I think we should point out in a comment that this is a first naive implementation and that you can just skip to end to get the streak.

kkuepper (changes requested):

web/src/components/EpisodeViewer.vue
sifferhans: This is essentially used to prevent spamming the endpoint.

Each tick, every ~10th of a second or so, the `onProgress` hook is called by the video player, and I check if the video progress is > 50%.

Each tick `playedOverHalfCounter` is incremented, and when the counter is exactly 10, I call the `/Statistics/watched` endpoint.
This makes sure we don't spam the endpoint, and the user has to keep playing for a second or so after rounding 50% for the endpoint to be called.

Overengineered?

web/src/components/EpisodeViewer.vue
```diff
@@ -250,6 +256,35 @@ const checkProgress = async (force?: boolean) => {
...
+let hasSentHalfwayWatchedEvent = false
+function onVideoProgress(
+    currentTime: number,
+    duration: number,
+    player: Player
+) {
+    const videoProgressPercent = currentTime / duration
+    if (
+        (videoProgressPercent >= 0.5 && !hasSentHalfwayWatchedEvent) ||
+        videoProgressPercent === 1
```
kkuepper, line 271: how often is this function called with videoProgressPercent === 1?
If it's more than once, we should make sure to send the event only once.

kkuepper (changes requested):

web/src/components/EpisodeViewer.vue
sifferhans: Based on my testing, it is called only once. When `currentTime === duration`, the player stops playing

kkuepper: ok. if you're confident in that this can be merged

kkuepper (approved):

web/src/components/EpisodeViewer.vue
kkuepper: it could make sense to add a comment though

PR #1221, opened 2025-06-26
Notify translators when there is new content to be translated

KillerX:


backend/translations/phrase/client.go
```diff
@@ -88,6 +88,8 @@ func (c *Client) SetDebug(debug bool) {
 
 type translationFile map[string]json.RawMessage
 
+const NotificationSourceChanged string = "HwiJ0yOI4knHriGQ531nP5"
```
sifferhans, line 91: Is this just hard-coded to make sure the job name is unique?

sifferhans: Ah, email template id. Could maybe name a bit better

sifferhans (approved):

PR #1224, opened 2025-07-04
Api/export v2

MasterMG2020:
introduces a new asynchronous export workflow via the ExportAsync API endpoint.

**Details:**
Adds ExportAsync to the API.

How It Works:
    Initial Request:
    When a user initiates an export via ExportAsync, the api creates a new entry in the users.export table in PG. Then it  responds immediately after with the newly generated exportId.

Polling:
The user can then use this exportId in the followup request to the same endpoint to check the status of the export.
Status Handling:
-     If the export is ready, the response includes the export URL and the status will be "ready".
-     If the export is not yet complete, the response will not have the URL and the status will be "new".


backend/loaders/loaders.go
```diff
@@ -0,0 +1,163 @@
+package loaders
```
KillerX, line 1: Rename file to batch_loaders

backend/loaders/loaders.go
```diff
@@ -0,0 +1,163 @@
...
+	PromptIDsLoader       func(ctx context.Context) ([]uuid.UUID, error)
+	FAQCategoryIDsLoader  func(ctx context.Context) ([]uuid.UUID, error)
+	ShortIDsLoader        func(ctx context.Context) ([][]uuid.UUID, error)
+	ShortWithScoresLoader func(ctx context.Context) ([]common.ShortIDWithMeta, error)
+}
+
+// PersonalizedLoaders contains loaders that are personalized to the user
+//
+// This includes things like permissions, settings, langauges, etc
+type PersonalizedLoaders struct {
```
KillerX, line 141: Move into separate file

backend/loaders/loaders.go
```diff
@@ -0,0 +1,163 @@
...
+	TimedMetadataLoader             *Loader[uuid.UUID, *common.TimedMetadata]
+	ChaptersLoader                  *Loader[int, []*common.TimedMetadata]
+	PersonLoader                    *Loader[uuid.UUID, *common.Person]
+	SongLoader                      *Loader[uuid.UUID, *common.Song]
+	PhraseLoader                    *Loader[string, *common.Phrase]
+	ContributionsLoader             *Loader[int32, *common.Contribution]
+}
+
+// LoadersWithPermissions contains loaders that will be filtered by permissions.
+type LoadersWithPermissions struct {
```
KillerX, line 104: Move into separate file

backend/loaders/conversion.go
```diff
@@ -2,28 +2,10 @@ package loaders
 
 import (
 	"context"
+	"github.com/bcc-code/bcc-media-platform/backend/common"
```
KillerX, line 5: Move the whole file into common package

backend/loaders/relation.go
```diff
@@ -2,35 +2,14 @@ package loaders
 
 import (
 	"context"
+	"github.com/bcc-code/bcc-media-platform/backend/common"
```
KillerX, line 5: Move the whole file into common package

migrations/00333_add_exports_table.sql
```diff
@@ -0,0 +1,37 @@
...
+    id uuid NOT NULL DEFAULT gen_random_uuid(),
+    profile_id uuid NOT NULL,
+    user_groups text[] NOT NULL DEFAULT '{}'::text[],
+    status text NOT NULL DEFAULT 'new',
+    created_date timestamp with time zone NOT NULL DEFAULT now(),
+    expiry_date timestamp with time zone,
+    url text NOT NULL DEFAULT '',
+    ContentOnlyInPreferredLanguage boolean NOT NULL DEFAULT false,
+    PreferredAudioLanguages text[] NOT NULL DEFAULT '{}'::text[],
+    PreferredSubtitlesLanguages text[] NOT NULL DEFAULT '{}'::text[],
```
KillerX, line 14: This should follow the same naming pattern as other fields.

migrations/00333_add_exports_table.sql
```diff
@@ -0,0 +1,37 @@
...
+    application_id integer NOT NULL DEFAULT 0,
+    application_code text NOT NULL DEFAULT '',
+    application_clientVersion text NOT NULL DEFAULT '',
+    application_default_page_id integer,
+
+    PRIMARY KEY (id),
+    CONSTRAINT fk_profile FOREIGN KEY (profile_id)
+        REFERENCES users.profiles (id) MATCH SIMPLE
+        ON UPDATE NO ACTION
+        ON DELETE NO ACTION
```
KillerX, line 25: On delete -> Cascacde

queries/exports.sql
```diff
@@ -0,0 +1,51 @@
...
+UPDATE users.exports
+SET url = @url
+WHERE id = @id::uuid;
+
+-- name: UpdateExportStatus :exec
+UPDATE users.exports
+SET status = @status
+WHERE id = @id::uuid;
+
+-- name: UPdateExpiryDate :exec
```
KillerX, line 48:
```suggestion
-- name: UpdateExpiryDate :exec
```

KillerX (changes requested):
Some mostly cosmetic changes that need to be done.

backend/loaders/relation.go
MasterMG2020: leaving this in loaders for now because of a circular dependency when importing loaders in common.

PR #1231, opened 2025-08-12
Add stacktrace rendering to GQL errors

KillerX:
The `fmt` package seems to have stripped the stacktrace from the error.


KillerX:
Ignore the changes in the helper file. They can be there but don't affect anything.

PR #1279, opened 2026-05-11
WEB: use actual links for navigation

sifferhans:
All episode cards etc. were just buttons previously, which worked against the default browser behaviors.
They are now changed to use actual link tags if prossible.

### DIFF
--- a/web/src/components/sections/item/CardSection.vue
+++ b/web/src/components/sections/item/CardSection.vue
@@ -20,7 +20,7 @@ const filteredItems = computed(() =>
     <section>
         <SectionTitle v-if="item.title">{{ item.title }}</SectionTitle>
         <div class="flex gap-4">
-            <div v-for="(i, index) in filteredItems">
+            <div v-for="(i, index) in filteredItems" :key="i.id">
                 <StudyTopicCardLarge
                     v-if="
                         item.cardSize == 'large' &&
--- a/web/src/components/sections/item/CardSection.vue
+++ b/web/src/components/sections/item/CardSection.vue
@@ -9,7 +9,7 @@ const props = defineProps<{
 }>()
 
 defineEmits<{
-    (event: 'clickItem', index: number): void
+    (event: 'clickItem', index: number, isModified: boolean): void
 }>()
 
 const filteredItems = computed(() =>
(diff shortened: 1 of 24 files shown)


web/src/components/sections/item/CollectionItemThumbnail.vue
```diff
@@ -14,28 +14,46 @@ import { useI18n } from 'vue-i18n'
...
+// the modifier-click case — listeners check the payload to decide whether to
+// also do SPA navigation.
+// eslint-disable-next-line no-undef
+const handleLinkClick = (event: MouseEvent) => {
+    const modified = isModifiedClick(event)
+    emit('click', modified)
+    if (modified) return
+    event.preventDefault()
+    clicked.value = true
 }
```
KillerX: This seems to be copy pasted several times. Could/should it be centralized?

web/src/components/sections/item/CollectionItemThumbnail.vue
```diff
@@ -14,28 +14,46 @@ import { useI18n } from 'vue-i18n'
...
     clicked.value = true
-    emit('click')
+    emit('click', false)
+}
+
+// On a real <a>, let modifier/middle clicks fall through so the browser opens
+// a new tab / window natively. We still emit `click` so analytics fires for
+// the modifier-click case — listeners check the payload to decide whether to
+// also do SPA navigation.
+// eslint-disable-next-line no-undef
```
KillerX: Claude suggests to instead `browser: true (or env: { browser: true }) to the ESLint config once, rather than suppressing per-file.`

Not sure if that is a valid way to go in this case but if that works it's imo more clean than the ignores.

KillerX:
Naming suggestion: consider renaming isModified / isModifiedClick to modifierKeyUsed. Makes it more clear what we are trying to track/do.

web/src/components/sections/item/CardSection.vue
```diff
@@ -28,7 +28,7 @@ const filteredItems = computed(() =>
                     "
                     ref="sectionItem"
                     :item="i.item"
-                    @click="$emit('clickItem', index)"
+                    @click="$emit('clickItem', index, false)"
```
KillerX, line 31: Do we intentionally force SPA navigation here?

KillerX (approved):

PR #1295, opened 2026-05-21
Feat/playlist section context

KillerX:
It is now possible to assign a playlist to a section. It should work more or less the same way as An Item section with a collection assigned.

GQL Also informs you about the context:

```
fragment ItemSection on ItemSection {
  metadata {
    collectionId
    playlistId
  }
}
```

```
"metadata": {
              "collectionId": "185",
              "playlistId": "109b78ac-8974-4485-8604-2c84176d6973",
            },
```

Context should also be ready to receive it.


sifferhans:
Nice, but is this only for basically having an inline view of the playlist? We should probably have a way to add links to the playlist page, right? Basically like in BMM

KillerX:
Right, that's of course missing too! Will add (waiting with merge until after the weekend.)

sifferhans (approved):

PR #1329, opened 2026-09-09
feat(message): add new message style variant enum to make it easier for apps to implement own design

sifferhans:


KillerX (approved):
Looks good. If I understand correctly, you are just reusing the existing field. 👍

